PR #7, opened 2019-09-23
SUBMARINE-195. Fix the incorrect project name in merge_submarine_pr.py

tangzhankun:
### What is this PR for?
When it merges the PR and the committer chose to close Jira, it fails with something like project "HADOOP-SUBMARINE" not be found. Tested that change it to "SUBMARINE" will work. 

### What type of PR is it?
Bug Fix
### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-195

use the cicd image to do the PR merge. And also choose yes to update the associated JIRA.
It will work with this fix.

### Questions:


tangzhankun:
Leave a message here to test if this comment will merge to jira.

PR #33, opened 2019-10-08
SUBMARINE-218. Update hadoop-3.2 profile to use hadoop 3.2.1.

jojochuang:
### What is this PR for?
A few sentences describing the overall goals of the pull request's commits.
Update hadoop-3.2 profile to use Apache Hadoop 3.2.1 release

Hadoop 3.2.1 updated guava version to 27.0-jre, which is not compatible with prior version, so I had to update guava version used too. httpcomponents was updated which differs from the one used by Tony, so exclude httpcomponents from Tony and use the version (newer in Hadoop 3.2) in Hadoop.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/SUBMARINE/

[Travis CI](https://travis-ci.org/jojochuang/hadoop-submarine/builds/594904607)

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,6 +229,7 @@
         <jackson-annotations.version>2.9.5</jackson-annotations.version>
         <guice.version>4.0</guice.version>
         <zookeeper.version>3.4.13</zookeeper.version>
+        <guava.version>27.0-jre</guava.version>
         <profile-id>hadoop-3.2</profile-id>
       </properties>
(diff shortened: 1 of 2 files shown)


yuanzac:
LGTM~
Thanks a lot. @jojochuang

PR #35, opened 2019-10-08
SUBMARINE-221. Update submarine version to 0.3.0-SNAPSHOT in mini-submarine.

yuanzac:
### What is this PR for?
The way to use submarine is quite different from the version 0.2.0. We need to update mini-submarine with the latest code so that users can try submarine more conveniently.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-221

https://travis-ci.org/yuanzac/hadoop-submarine/builds/595068622

### Questions:
* Is there breaking changes for older versions? Yes

### DIFF
--- a/dev-support/mini-submarine/conf/submarine.xml
+++ b/dev-support/mini-submarine/conf/submarine.xml
@@ -19,7 +19,7 @@
 <configuration>
   <property>
     <name>submarine.runtime.class</name>
-    <value>org.apache.hadoop.yarn.submarine.runtimes.tony.TonyRuntimeFactory</value>
+    <value>org.apache.submarine.runtimes.tony.TonyRuntimeFactory</value>
     <!--
     <value>org.apache.hadoop.yarn.submarine.runtimes.yarnservice.YarnServiceRuntimeFactory</value>
     -->
--- a/dev-support/mini-submarine/conf/submarine.xml
+++ b/dev-support/mini-submarine/conf/submarine.xml
@@ -21,7 +21,7 @@
     <name>submarine.runtime.class</name>
     <value>org.apache.submarine.runtimes.tony.TonyRuntimeFactory</value>
     <!--
-    <value>org.apache.hadoop.yarn.submarine.runtimes.yarnservice.YarnServiceRuntimeFactory</value>
+    <value>org.apache.submarine.runtimes.yarnservice.YarnServiceRuntimeFactory</value>
     -->
   </property>
 </configuration>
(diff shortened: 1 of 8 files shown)


dev-support/mini-submarine/conf/submarine.xml
```diff
@@ -19,7 +19,7 @@
 <configuration>
   <property>
     <name>submarine.runtime.class</name>
-    <value>org.apache.hadoop.yarn.submarine.runtimes.tony.TonyRuntimeFactory</value>
+    <value>org.apache.submarine.runtimes.tony.TonyRuntimeFactory</value>
     <!--
     <value>org.apache.hadoop.yarn.submarine.runtimes.yarnservice.YarnServiceRuntimeFactory</value>
     -->
```
pingsutw: we also need to change it to **org.apache.submarine.runtimes.yarnservice.YarnServiceRuntimeFactory**

pom.xml
```diff
@@ -185,6 +185,11 @@
         <artifactId>jackson-annotations</artifactId>
         <version>${jackson-annotations.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
```
liuxunorg: Why do you need to import `jackson` jar?

submarine-core/pom.xml
```diff
@@ -124,6 +124,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
```
liuxunorg: Why do you need to import `jackson` jar?

dev-support/mini-submarine/conf/submarine.xml
yuanzac: It makes sense to me.
Thanks, @pingsutw

pom.xml
yuanzac: I just found a bug that, if we don't import jackson jar, we would get the error that jackson class is not found, when using submarine-all-*.jar to submit a submarine job.

submarine-core/pom.xml
yuanzac: The error that jackson class is not found if jackson-core is not imported

liuxunorg:
@yuanzac , PR title loss word, `in mini-sub…`

yuanzac:
> @yuanzac , PR title loss word, `in mini-sub…`

The title of this PR is changed

yuanzac:
I'll merge this pr later today, if there are no other comments.

PR #36, opened 2019-10-08
SUBMARINE-222. Use virtual Python env in CICD image

tangzhankun:
### What is this PR for?
When using the current CICD image, there is a possibility that the merge PR script will fail due to Python environment issue. I propose to use python virtual env to solve this.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-222

re-build the cicd image and use it to merge PR
### Questions:

### DIFF
--- a/dev-support/cicd/Dockerfile
+++ b/dev-support/cicd/Dockerfile
@@ -16,8 +16,16 @@
 FROM ubuntu:16.04
 RUN \
   apt-get update && \
-  apt-get install -y vim python git python-pip && \
-  pip install jira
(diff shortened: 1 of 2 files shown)


jojochuang:
Looks good.
Instead of downloading virtualenv source code & compile it, you should be able to just install it: as an example: https://gist.github.com/frfahim/73c0fad6350332cef7a653bcd762f08d

yuanzac:
LGTM.
I've tested it. it works. 
Thanks @tangzhankun @jojochuang @liuxunorg

PR #40, opened 2019-10-10
SUBMARINE-229. Submarine-cicd failed due to the encoding error of github PR body

yuanzac:
### What is this PR for?
Sometimes, submarine-cicd failed to merge pr.
The root cause is that the encode of PR body is UTF-8, which would cause python str encoding error.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-229

https://travis-ci.org/yuanzac/hadoop-submarine/builds/596113339

### Questions:

### DIFF
--- a/dev-support/cicd/README.md
+++ b/dev-support/cicd/README.md
@@ -21,25 +21,31 @@ To use them more easily, we provide a Docker image to help committer to handle t
 ## Docker mode
 
 ```
-cd <path-to-submarine-home>/dev-support/cicd
-docker build -t submarine-cicd .
-docker run -it --rm submarine-cicd
+./build_and_start_cicd_image.sh
 ```
 
 Or
 
 ```
-./build_and_start_cicd_image.sh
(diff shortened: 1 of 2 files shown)


dev-support/cicd/README.md
```diff
@@ -21,25 +21,30 @@ To use them more easily, we provide a Docker image to help committer to handle t
...
+docker run -it --rm submarine-cicd
 ```
 
+Apache jira username and password are required in the docker container. You can
+add them to the docker startup command.
+
 And you'll see output like below and then you can decide what to accomplish.
 ```
-$ docker run -it --rm submarine-cicd
+$ docker run -it -e JIRA_USERNAME='username' -e JIRA_PASSWORD='password' --rm submarine-cicd
```
tangzhankun: it can also set these environment variable "-e APACHE_ID=aid -e APACHE_NAME="Zhankun Tang""

yuanzac: Done

PR #45, opened 2019-10-14
[SUBMARINE-237] Fixed workbench-daemon.sh returns wrong status

luzhonghao:
### What is this PR for?
We start workbench server successfully. However, when  run "workbench-daemon.sh status"  , it returns message "Submarine Workbench is not running" 

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-237

manual test 
* start workbench server
* run “workbench-daemon.sh  status ” command and see if the return message is correct

### Questions:

### DIFF
--- a/bin/workbench-daemon.sh
+++ b/bin/workbench-daemon.sh
@@ -148,15 +148,15 @@ function find_workbench_process() {
   pid=`found_workbench_server_pid`
 
   if [[ -z "$pid" ]]; then
+    echo "${WORKBENCH_NAME} is not running"
+    return 1
+  else
     if ! kill -0 ${pid} > /dev/null 2>&1; then
       echo "${WORKBENCH_NAME} running but process is dead"
       return 1
     else
       echo "${WORKBENCH_NAME} is running"
     fi
-  else
-    echo "${WORKBENCH_NAME} is not running"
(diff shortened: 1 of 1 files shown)


bin/workbench-daemon.sh
```diff
@@ -148,15 +148,15 @@ function find_workbench_process() {
   pid=`found_workbench_server_pid`
 
   if [[ -z "$pid" ]]; then
-    if ! kill -0 ${pid} > /dev/null 2>&1; then
+    echo "${WORKBENCH_NAME} is not running"
+    return 1
+  else
+   if ! kill -0 ${pid} > /dev/null 2>&1; then
```
liuxunorg: The first letter needs to be indented into two.

liuxunorg:
Can you shorten the title of the PR?

luzhonghao:
> Can you shorten the title of the PR?



> Can you shorten the title of the PR?

Done, thanks for review

bin/workbench-daemon.sh
luzhonghao: > The first letter needs to be indented into two.

Done, thanks to point it

PR #38, opened 2019-10-09
[SUBMARINE-220] Update Java dependency.

jojochuang:
### What is this PR for?
Update java dependencies with known CVEs.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-220

### How should this be tested?
[Travis CI](https://travis-ci.org/jojochuang/hadoop-submarine/builds/595393903)

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #49, opened 2019-10-15
SUBMARINE-243. Workbench failed to save project.

yuanzac:
### What is this PR for?
When we save a project in workbench, we get the following error:

javax.servlet.ServletException: org.glassfish.jersey.server.ContainerException: java.lang.NoClassDefFoundError: com/fasterxml/jackson/annotation/JsonMerge

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-243

https://travis-ci.org/yuanzac/hadoop-submarine/builds/598019153

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,7 +229,6 @@
         <jaxb-api.version>2.2.11</jaxb-api.version>
         <commons-compress.version>1.19</commons-compress.version>
         <guice-servlet.version>4.0</guice-servlet.version>
-        <jackson-annotations.version>2.9.5</jackson-annotations.version>
         <guice.version>4.0</guice.version>
         <zookeeper.version>3.4.13</zookeeper.version>
         <guava.version>27.0-jre</guava.version>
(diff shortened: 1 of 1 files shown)


yuanzac:
Due to the error:
java.lang.NoClassDefFoundError: com/fasterxml/jackson/annotation/JsonMerge
We need to use the jackson-annotations jar of which the version is 2.9.
Since PR #38 has set the following attribute in the pom.xml:
<jackson-annotations.version>2.9.10</jackson-annotations.version>
We need to delete the variable, jackson-annotations.version, from the profiles to make PR #38 take effects.

yuanzac:
Thanks @liuxunorg for the review~
There already is jackson-annotations.version definition in the previous part of pom.xml. It is in line 66 of the pom.xml.
Deleting all jackson-annotations.jar in the profiles of pom.xml would make line 66 take effect.

liuxunorg:
Will merge if no more comments

PR #75, opened 2019-10-29
SUBMARINE-268. Unify all test class file name with "*Test" pattern

hhhizzz:
### What is this PR for?
Due to history reasons, we have the test class file named with "Test*" mainly in submarine-client.
In order to unify this, I propose that we use "Test" as the suffix in the file name. Then we need to rename all needed test files.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-268?filter=allopenissues

https://travis-ci.org/hhhizzz/hadoop-submarine/builds/604332888

### Questions:
* Does the licenses files need an update? No
* Are there breaking changes for older versions? Yes，The name of Tests has been changed


yuanzac:
LGTM,
Thanks @hhhizzz

PR #77, opened 2019-11-01
SUBMARINE-272. Modules build order is incorrect.

yuanzac:
### What is this PR for?
Submarine dist module is used to build a submarine code package. It is supposed to contains all the submarine modules, including workbench and submarine server. 
But maven build order is incorrect, which would cause that workbench code is missing.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-272

https://travis-ci.org/yuanzac/hadoop-submarine/builds/605900900?utm_medium=notification&utm_source=github_status

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,14 +117,14 @@
   </properties>
 
   <modules>
-    <module>submarine-all</module>
     <module>submarine-client</module>
     <module>submarine-commons</module>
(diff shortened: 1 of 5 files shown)


yuanzac:
Thanks @jiwq for the review~

PR #83, opened 2019-11-06
SUBMARINE-280. Reorganize dependencies in modules

LiShuMing:
### What is this PR for?
Now a lot of modules are using dependencies in its self pom.xml separately which is redundant and messy, so this pr reorganizes dependencies of sub-module so that they depend the parent dependencies management.

### What type of PR is it?
[Improvement]

### Todos
* Fix compile warnings.
- Fix plugins dependencies.

### What is the Jira issue?
* https://issues.apache.org/jira/browse/SUBMARINE-280

### Questions:


pom.xml
```diff
@@ -240,6 +205,455 @@
...
+        <artifactId>gson</artifactId>
+        <version>${gson.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>commons-logging</groupId>
+        <artifactId>commons-logging</artifactId>
+        <version>${commons.logging.version}</version>
+      </dependency>
+
+
```
jiwq: redundant blank line

submarine-server/server-submitter/submitter-yarnservice/pom.xml
```diff
@@ -26,6 +26,7 @@
     <groupId>org.apache.submarine</groupId>
     <artifactId>server-submitter</artifactId>
     <version>0.3.0-SNAPSHOT</version>
+    <relativePath>../../../pom.xml</relativePath>
```
jiwq: wrong path

pom.xml
LiShuMing: OK， thanks for your reviews.

LiShuMing:
cc @jiwq I've resolved all the CI failures. Can you help me review codes again?

submarine-workbench/interpreter/interpreter-engine/pom.xml
```diff
@@ -27,7 +27,7 @@
     <groupId>org.apache.submarine</groupId>
     <artifactId>interpreter</artifactId>
     <version>0.3.0-SNAPSHOT</version>
-    <relativePath>../pom.xml</relativePath>
+    <relativePath>..</relativePath>
```
yuanzac: Why do we need to change "../pom.xml" to ".."?
../pom.xml is default value of relativePath

pom.xml
```diff
@@ -240,6 +205,786 @@
...
+      <dependency>
+        <groupId>com.github.pagehelper</groupId>
+        <artifactId>pagehelper</artifactId>
+        <version>${pagehelper.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.glassfish.jersey.core</groupId>
+        <artifactId>jersey-client</artifactId>
+        <version>${jersey.version}</version>
+        <exclusions>
```
yuanzac: In my opinion, We'd better not add <exclusions> to the <dependencyManagement>. Because it's only useful if there are dependency conflictions.

PR #87, opened 2019-11-08
SUBMARINE-267. Initial job server which defines REST skeleton

tangzhankun:
### What is this PR for?
The initial job server which defines REST skeleton

### What type of PR is it?
Feature 

### Todos
* [1] - Fix the YAML unit test case issue
* [2] - Populate the REST API with integration to yarn-submitter or k8s-submitter
* [3] - Add Swagger to easy access the API spec

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-267

CI unit test is enough for the REST API before we integrate with submitter

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
   global:
     # submarine core does not required by workbench-server integration tests
     # If you need to compile Phadoop-3.1 or Phadoop-3.2, you need to add `!submarine-server/server-submitter/submitter-yarnservice` in EXCLUDE_SUBMARINE
-    - EXCLUDE_SUBMARINE="!submarine-all,!submarine-client,!submarine-commons,!submarine-commons/commons-runtime,!submarine-dist,!submarine-server/server-submitter/submitter-yarn,!submarine-server/server-submitter/submitter-k8s"
+    - EXCLUDE_SUBMARINE="!submarine-all,!submarine-client,!submarine-commons,!submarine-commons/commons-runtime,!submarine-dist,!submarine-server/server-submitter/submitter-yarn,!submarine-server/server-submitter/submitter-k8s,!submarine-server/server-core"
     - EXCLUDE_WORKBENCH="!submarine-workbench,!submarine-workbench/workbench-web,!submarine-workbench/workbench-server"
     - EXCLUDE_INTERPRETER="!submarine-workbench/interpreter,!submarine-workbench/interpreter/interpreter-engine,!submarine-workbench/interpreter/python-interpreter,!submarine-workbench/interpreter/spark-interpreter""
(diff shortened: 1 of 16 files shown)


submarine-server/server-core/src/test/java/JobApiTest.java
```diff
@@ -0,0 +1,146 @@
...
+        .request()
+        .post(Entity.entity(jobSpec, MediaType.APPLICATION_JSON));
+
+    assertEquals("Http Response should be 202 ",
+        Response.Status.ACCEPTED.getStatusCode(), response.getStatus());
+  }
+
+  // Test job created with incorrect JSON input
+  @Test
+  public void testCreateJobWhenJsonInputIsWrongThenResponseCodeAccepted() {
```
pingsutw: would it be better to change to `testCreateJobWhenJsonInputIsWrongThenResponseCodeBadRequest` ?

yuanzac:
@tangzhankun, Thanks for your contribution.
The test is failed. Can you take a look?

submarine-server/server-core/src/test/java/JobApiTest.java
tangzhankun: Good catch. Will change it.

submarine-server/server-core/src/main/java/org/apache/submarine/jobserver/JobServer.java
```diff
@@ -0,0 +1,165 @@
...
+    apiServlet.setInitParameter("jersey.config.server.provider.packages",
+        "org.apache.submarine.jobserver.rest");
+
+    try {
+      jobServer.start();
+      LOG.info("Submarine job server started");
+      jobServer.join();
+    } catch (Exception e) {
+      LOG.info("Submarine job server failed to start");
+      e.printStackTrace();
```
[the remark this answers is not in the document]
tangzhankun: Good catch! Will change it.

tangzhankun:
@yuanzac @jiwq , my local build doesn't have this dependency convergence issue. Could you please help to clone the code and "mvn clean package -DskipTests"?

.travis.yml
```diff
@@ -160,6 +160,13 @@ matrix:
         - npm run e2e -- --protractor-config=e2e/protractor-ci.conf.js
       env: NAME="Build workbench-web-ng"
 
+
+    # Test submarine-server
+    - language: java
+      jdk: "openjdk8"
+      dist: xenial
+      env: NAME="Test submarine-server" PROFILE="-Phadoop-2.9" BUILD_FLAG="clean package install -DskipTests" TEST_FLAG="test -DskipRat -am" MODULES="-pl ${EXCLUDE_WORKBENCH},${EXCLUDE_INTERPRETER}" TEST_MODULES="-pl submarine-server/server-core" TEST_PROJECTS=""
```
yuanzac: There is another test named "Test submarine-server" in line 107 as well. 
It looks like their functionalities are the same. Can we combine them?

liuxunorg: @yuanzac line 107 is "Test workbench-server". Not is "Test submarine-server".

PR #98, opened 2019-11-21
SUBMARINE-293. Refactoring submarine server rest test framework

liuxunorg:
### What is this PR for?
Provides a way to actually run the submarin server REST service and test the rest interface in the test case.
This provides a realistic runtime environment to test whether each REST interface is working correctly.

### What type of PR is it?
[Refactoring]

### What is the Jira issue?
* https://issues.apache.org/jira/browse/SUBMARINE-293

* [CI Pass](https://travis-ci.org/liuxunorg/submarine/builds/615024635)

### Questions:


submarine-server/server-core/src/main/java/org/apache/submarine/server/utils/GitRequestUtil.java
```diff
@@ -32,8 +32,8 @@
 import java.util.Date;
 import java.util.Map;
 
-public class HttpRequestUtil {
-  private static final Logger LOG = LoggerFactory.getLogger(HttpRequestUtil.class);
+public class GitRequestUtil {
```
yuanzac: Do we need to change HttpRequestUtil to GitRequestUtil?

liuxunorg: HttpRequestUtil is a special class for git http request, Change to GitHttpRequest.

PR #108, opened 2019-11-26
SUBMARINE-302. Split the submitter-yarnservice from travis into job

jiwq:
### What is this PR for?
The submitter-yarnservice depend on hadoop (3.1+), but it always break other modules. So we'd better split it into job from travis at first.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-302

https://travis-ci.org/jiwq/submarine/builds/617271191

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 sudo: required
+dist: xenial
+language: java
 
 before_cache:
   - sudo chown -R travis:travis $HOME/.m2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -167,6 +154,18 @@ matrix:
       dist: xenial
       env: NAME="Test submarine-server" PROFILE="-Phadoop-2.9" BUILD_FLAG="clean package install -DskipTests" TEST_FLAG="test -DskipRat -am" MODULES="-pl ${EXCLUDE_K8S},${EXCLUDE_WORKBENCH},${EXCLUDE_INTERPRETER}" TEST_MODULES="-pl submarine-server/server-core" TEST_PROJECTS=""
 
+    - name: Submarine on Hadoop YARN 3.1
```
yuanzac: In my opinion, we can't just test yarn server for hadoop 3.1 and hadoop3.2. Submarine other modules, even tony has compatible issues for hadoop 3.*. Because hadoop API is different between hadoop 2 and hadoop 3.
If we want to separate yarn service, we can add a new travis job for yarn service just like this. But we should keep test job for other modules.

jiwq: Sure, you are right. At first we split the submitter-yarnservice to ensure the other module's (such as  meta store) work can continue. 

Ideally, submitter-yarn and submitter-yarnservice belongs to the same job, so we should deprecated the submarine-all and submarine-dist when the server passed the POC. Then we can merge the yarn and yarnservice to one job, now only the yarn related module depend on the Hadoop.

yuanzac: @jiwq, Thanks for your comments.
The issue of meta store will be fixed by Liuxun soon. It is not related with yarnservice or tony.
I agree to deprecate yarnservice. But we should keep tests for hadoop3. Because if we add hadoop3 test later, some PR maybe umcompatible with hadoop 3. It will acquire a lot extra efforts to fix that.

jiwq: Yeah, maybe we should split the hadoop related module to the same job in travis module. Do you know which modules depend on hadoop? 
In my opinion, we should convert the external ecosystem related module as the plugin in future, so that the submarine server can support different version at the same time. From the operators perspective, we should do it. Any thoughts?

jojochuang: Are we good to resolve this comment? Looks to me we have agreed to resolve and will raise new jira for additional improvement.

jiwq: @jojochuang Thanks for the reply, I will fix the conflicts and if no more comments it can be merged.

jiwq:
Nowadays it's not the key feature, I will close it. If we need it, I will reopen this PR.

PR #117, opened 2019-11-29
SUBMARINE-303. Submarine client submit job to submarine server by grpc.

yuanzac:
### What is this PR for?
Submarine server is supposed to manage the job's lifecycle. Submarine client can submit a job to submarine server by using grpc, which is suitable for bi-direction communications.
A new configuration, named submarine.server.remote.execution.enabled, is added. If it is true,
submarine client will submit a job to submarine server instead of running a job by itself. 

### What type of PR is it?
Feature

### Todos
* - Need to add UT in the next PR. 

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-303

https://travis-ci.org/yuanzac/hadoop-submarine/builds/618607014?utm_source=github_status&utm_medium=notification

### Questions:

### DIFF
--- a/bin/submarine-daemon.sh
+++ b/bin/submarine-daemon.sh
@@ -36,9 +36,9 @@ cd ${BIN}/>/dev/null
 SUBMARINE_SERVER_NAME="Submarine Server"
 SUBMARINE_SERVER_LOGFILE="${SUBMARINE_LOG_DIR}/submarine.log"
 SUBMARINE_SERVER_MAIN=org.apache.submarine.server.SubmarineServer
-JAVA_OPTS+=" -Dsubmarine.log.file=${SUBMARINE_SERVER_LOGFILE}"
+JAVA_OPTS+="${SUBMARINE_SERVER_JAVA_OPTS} ${SUBMARINE_SERVER_MEM} -Dsubmarine.log.file=${SUBMARINE_SERVER_LOGFILE}"
 
-add_jar_in_dir "${BIN}/../lib"
+add_each_jar_in_dir_recursive "${BIN}/../lib"
 
 function initialize_default_directories() {
   if [[ ! -d "${SUBMARINE_LOG_DIR}" ]]; then
--- a/bin/submarine-daemon.sh
+++ b/bin/submarine-daemon.sh
@@ -38,7 +38,8 @@ SUBMARINE_SERVER_LOGFILE="${SUBMARINE_LOG_DIR}/submarine.log"
 SUBMARINE_SERVER_MAIN=org.apache.submarine.server.SubmarineServer
 JAVA_OPTS+="${SUBMARINE_SERVER_JAVA_OPTS} ${SUBMARINE_SERVER_MEM} -Dsubmarine.log.file=${SUBMARINE_SERVER_LOGFILE}"
 
-add_each_jar_in_dir_recursive "${BIN}/../lib"
+add_each_jar_in_dir "${BIN}/../lib"
+add_each_jar_in_dir "${BIN}/../lib/submitter"
 
 function initialize_default_directories() {
   if [[ ! -d "${SUBMARINE_LOG_DIR}" ]]; then
--- a/dev-support/mini-submarine/conf/bootstrap.sh
+++ b/dev-support/mini-submarine/conf/bootstrap.sh
@@ -27,9 +27,19 @@ CONFIG_DIR="/tmp/hadoop-config"
 
 # Copy config files from volume mount
 
-for f in slaves core-site.xml hdfs-site.xml mapred-site.xml yarn-site.xml container-executor.cfg capacity-scheduler.xml node-resources.xml resource-types.xml submarine-site.xml; do
+for f in slaves core-site.xml hdfs-site.xml mapred-site.xml yarn-site.xml container-executor.cfg capacity-scheduler.xml node-resources.xml resource-types.xml; do
   if [[ -e ${CONFIG_DIR}/$f ]]; then
     cp ${CONFIG_DIR}/$f $HADOOP_PREFIX/etc/hadoop/$f
+  else
+    echo "ERROR: Could not find $f in $CONFIG_DIR"
+    exit 1
+  fi
+done
+
+# Copy submarine config
+for f in slaves submarine-site.xml submarine-env.sh; do
+  if [[ -e ${CONFIG_DIR}/$f ]]; then
+    cp ${CONFIG_DIR}/$f /opt/submarine-current/conf/$f
   else
     echo "ERROR: Could not find $f in $CONFIG_DIR"
     exit 1
--- a/submarine-client/src/main/java/org/apache/submarine/client/cli/param/Quicklink.java
+++ b/submarine-client/src/main/java/org/apache/submarine/client/cli/param/Quicklink.java
@@ -62,15 +62,35 @@ public class Quicklink {
     return label;
   }
 
+  public Quicklink setLabel(String label) {
+    this.label = label;
+    return this;
+  }
+
   public String getComponentInstanceName() {
     return componentInstanceName;
+  }
+
+  public Quicklink setComponentInstanceName(String componentInstanceName) {
+    this.componentInstanceName = componentInstanceName;
+    return this;
   }
 
   public String getProtocol() {
     return protocol;
   }
 
+  public Quicklink setProtocol(String protocol) {
+    this.protocol = protocol;
+    return this;
+  }
+
   public int getPort() {
     return port;
+  }
+
+  public Quicklink setPort(int port) {
+    this.port = port;
+    return this;
   }
 }
--- a/submarine-commons/commons-rpc/src/main/proto/SubmarineServerProtocol.proto
+++ b/submarine-commons/commons-rpc/src/main/proto/SubmarineServerProtocol.proto
@@ -1,0 +1,129 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+
+//   http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_package = "org.apache.submarine.commons.rpc";
+option java_outer_classname = "SubmarineServerRpc";
+
+// Interface exported by the server.
+service SubmarineServerProtocol {
+  // Submit a job to submarine server.
+  rpc SubmitJob(ParameterProto) returns (ApplicationIdProto) {}
+
+  // test rpc
+  rpc TestRpc(ParametersHolderProto) returns (ApplicationIdProto) {}
+
+  // A server-to-client streaming RPC.
+  //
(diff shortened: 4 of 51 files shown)


dev-support/mini-submarine/conf/bootstrap.sh
```diff
@@ -36,6 +36,16 @@ for f in slaves core-site.xml hdfs-site.xml mapred-site.xml yarn-site.xml contai
   fi
 done
 
+# Copy submarine config
+for f in slaves submarine-site.xml submarine-env.sh; do
+  if [[ -e ${CONFIG_DIR}/$f ]]; then
+    cp ${CONFIG_DIR}/$f /opt/submarine-current/conf/$f
```
pingsutw: `/opt/submarine-current/conf/`  already has latest submarine config, I think we could remove **submarine-site.xml** and **submarine-env.sh** in `dev-support/mini-submarine/conf/`, so when submarine config change we don't need to update config in mini-submarine as well

submarine-server/server-rpc/src/test/java/org/apache/submarine/server/rpc/SubmarineServerClient.java
```diff
@@ -0,0 +1,89 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+
+/**
+ * Sample client code that makes gRPC calls to the server.
+ */
+public class SubmarineServerClient {
```
pingsutw: Would it be better if it changes to `SubmarineRpcClient`

dev-support/mini-submarine/conf/bootstrap.sh
yuanzac: @pingsutw, Thanks for the review.
/opt/submarine-current/conf/ has the latest submarine-site.xml and submarine-env.sh with default configuration. So we need to update them with the configuration suitable for mini-submarine.

submarine-server/server-rpc/src/test/java/org/apache/submarine/server/rpc/SubmarineServerClient.java
yuanzac: Thanks. make sense to me.

submarine-workbench/interpreter/spark-interpreter/pom.xml
```diff
@@ -175,6 +175,7 @@
       <groupId>org.apache.zeppelin</groupId>
       <artifactId>zeppelin-spark-dependencies</artifactId>
       <version>${zeppelin.version}</version>
+      <!--
       <exclusions>
```
liuxunorg: @hhhizzz Please help review this code, Thanks!

bin/submarine-daemon.sh
```diff
@@ -36,9 +36,9 @@ cd ${BIN}/>/dev/null
 SUBMARINE_SERVER_NAME="Submarine Server"
 SUBMARINE_SERVER_LOGFILE="${SUBMARINE_LOG_DIR}/submarine.log"
 SUBMARINE_SERVER_MAIN=org.apache.submarine.server.SubmarineServer
-JAVA_OPTS+=" -Dsubmarine.log.file=${SUBMARINE_SERVER_LOGFILE}"
+JAVA_OPTS+="${SUBMARINE_SERVER_JAVA_OPTS} ${SUBMARINE_SERVER_MEM} -Dsubmarine.log.file=${SUBMARINE_SERVER_LOGFILE}"
 
-add_jar_in_dir "${BIN}/../lib"
+add_each_jar_in_dir_recursive "${BIN}/../lib"
```
hhhizzz: maybe create a variable SUBMARINE_HOME=${BIN}/.. and then add_each_jar_in_dir_recursive "${SUBMARINE_HOME}/lib"  is better?

submarine-workbench/interpreter/spark-interpreter/pom.xml
yuanzac: Thanks @liuxunorg,
Just tried to figure out how to pass CI tests. I triggered travis for my repo. And it triggered this PR as well.
I'm still working on adding UT tests for this PR. This will be fixed shortly.

bin/submarine-daemon.sh
liuxunorg: `SUBMARINE_HOME` already exist in `conf/submarine-env.sh`

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
```diff
@@ -0,0 +1,407 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A sample gRPC server that serve the RouteGuide (see route_guide.proto) service.
```
tangzhankun: Could you please refine the comments? I don't see the "route_guide.proto" in this PR.

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
```diff
@@ -0,0 +1,407 @@
...
+        SubmarineConfiguration.getInstance();
+    int rpcServerPort = submarineConfiguration.getInt(
+        SubmarineConfiguration.ConfVars.SUBMARINE_SERVER_REMOTE_EXECUTION_PORT);
+    SubmarineRpcServer server = new SubmarineRpcServer(rpcServerPort);
+    server.start();
+    return server;
+  }
+
+  /**
+   * Our implementation of RouteGuide service.
```
tangzhankun: This comment seems confusing to me also.

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
```diff
@@ -0,0 +1,407 @@
...
+      roleParameters.setLaunchCommand(roleParameterProto.getLaunchCommand());
+    }
+
+    roleParameters.setReplicas(roleParameterProto.getReplicas())
+        .setResource(convertResourceProtoToResource(
+            roleParameterProto.getResourceProto()));
+    return roleParameters;
+  }
+
+  private static List<Localization> convertLocalizationProtoToLocalization(
```
tangzhankun: I'm thinking if the localization parameter with a "local:///home/user/foo/test.py:./test.py" would work in RPC server side. It seems the first local path can only be remote url like hdfs or S3.etc.

submarine-commons/commons-rpc/src/main/proto/SubmarineServerProtocol.proto
```diff
@@ -0,0 +1,129 @@
...
+  // Accepts a stream of RouteNotes sent while a route is being traversed,
+  // while receiving other RouteNotes (e.g. from other users).
+  // rpc RouteChat(stream RouteNote) returns (stream RouteNote) {}
+}
+
+message ParametersHolderProto {
+  int32 helloworld = 1;
+}
+
+message ParameterProto {
```
tangzhankun: I was thinking what's the relationship between this parameter proto message and the submarine server REST job spec. Would it be a good idea to use this proto as a source of truth for all jobs? Which means, the current REST job spec will be used by the workbench, and when submarine client submit job with RPC protocal, the job spec will be translated to this ParameterProto. When submarine client submit job with REST protocal, the job spec will be sent directly to submarine server.

submarine-workbench/interpreter/spark-interpreter/pom.xml
yuanzac: Done

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
yuanzac: Done.

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
yuanzac: Done

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
yuanzac: Yeah, The localization like "local:///home/user/foo/test.py:./test.py" would not work.

bin/submarine-daemon.sh
yuanzac: For now, let's just use ${BIN}/../lib, as I'm not sure about the conflicts with SUBMARINE_HOME in submarine-env.sh
@hhhizzz, If you have the bandwidth, help us to optimize it~

submarine-commons/commons-rpc/src/main/proto/SubmarineServerProtocol.proto
liuxunorg: Yes, we currently need two protocols to support the workbench and client respectively. 
My suggestion is not to stop the action and continue to promote the development of the project. 
We will reconsider this issue later.

yuanzac:
Add UT for RPC Server.

submarine-client/src/main/java/org/apache/submarine/client/cli/param/Quicklink.java
```diff
@@ -62,15 +62,35 @@ public String getLabel() {
     return label;
   }
 
+  public Quicklink setLabel(String label) {
+    this.label = label;
+    return this;
```
jiwq: Setter should not return a value.

submarine-client/src/main/java/org/apache/submarine/client/cli/param/Quicklink.java
```diff
@@ -62,15 +62,35 @@ public String getLabel() {
...
+  }
+
   public String getComponentInstanceName() {
     return componentInstanceName;
   }
 
+  public Quicklink setComponentInstanceName(String componentInstanceName) {
+    this.componentInstanceName = componentInstanceName;
+    return this;
```
jiwq: Setter should not return a value.

submarine-client/src/main/java/org/apache/submarine/client/cli/param/Quicklink.java
```diff
@@ -62,15 +62,35 @@ public String getLabel() {
...
+  }
+
   public String getProtocol() {
     return protocol;
   }
 
+  public Quicklink setProtocol(String protocol) {
+    this.protocol = protocol;
+    return this;
```
jiwq: Setter should not return a value.

submarine-client/src/main/java/org/apache/submarine/client/cli/param/Quicklink.java
```diff
@@ -62,15 +62,35 @@ public String getLabel() {
...
+  }
+
   public int getPort() {
     return port;
   }
+
+  public Quicklink setPort(int port) {
+    this.port = port;
+    return this;
```
jiwq: Setter should not return a value.

yuanzac:
@jiwq
Thanks a lot for the review~

1. Setter should not return a value.
It is used for SubmarineRpcServerProto and ClientProto. Just like a builder, with a return value.
we can make the conversion between JOBO and protobuf object more concisely.

2. Should create the common-rpc and server-rpc module? In my opinion, it is not required. The RPC server belongs to the submarine server.
Since commons-rpc depends on protobuf 3.*, which is incompatible with hadoop protobuf 2.*. We'd better use a seperate module and shade the dependency to avoid jar conflicts.

3. The PB is related to the Generic Job Spec, you'd better create a JIRA ticket to tracking this.
This PR doesn't make Generic Job Spec modifications. Zhankun is working on it and it'll be a seperated JIRA.

yuanzac:
There are still some jar conflicts, I'll fix it shortly.

PR #119, opened 2019-12-06
SUBMARINE-310. Support golang and java mixed compilation

liuxunorg:
### What is this PR for?
Submarine needs to support both the yarn and k8s resource scheduling systems for job operation.
Some modules (`submitter-yarn`) are developed using java, and some modules (`submarine-operator` / `submitter-k8s`) are developed using golang.
So submarine needs to support both `mvn` and `go build` compilation.

cd submarine path, execute :
```
mvn package
```
or
```
go build
```

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/SUBMARINE-310

* [CI Pass](https://travis-ci.org/liuxunorg/submarine/builds/621674402)

#### mvn package

image

#### go build

image

### Questions:


liuxunorg:
After discussion with the community, keep the original module structure unchanged, add the `submarine-cloud` folder to the project root directory, all `submarine-cloud` folders are developed using the `Go language` to support submarine on k8s (aws, gce, azure in the future).

Therefore, this PR will be closed.

PR #130, opened 2019-12-17
SUBMARINE-321. Add JobManager and SubmitterManager components

jiwq:
### What is this PR for?
The JobManager is responsible for cache and schedule the job to submitter.
The SubmitterManager help to load the different submitter plugins.

### What type of PR is it?
[ Feature ]

### Todos
* [ ] - Refactor submitter

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-321

### Questions:

### DIFF
--- a/submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfiguration.java
+++ b/submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfiguration.java
@@ -28,9 +28,11 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.StringTokenizer;
 
 public class SubmarineConfiguration extends XMLConfiguration {
   private static final Logger LOG = LoggerFactory.getLogger(SubmarineConfiguration.class);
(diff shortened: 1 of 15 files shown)


submarine-server/server-core/src/main/java/org/apache/submarine/server/JobManager.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class JobManager {
+  private static final Logger LOG = LoggerFactory.getLogger(JobManager.class);
+  private static JobManager manager;
```
tangzhankun: This manager should be "private static volatile". The volatile is a must for double check locking per my understanding.

tangzhankun:
Thanks a lot! Wanqiang!
The patch looks good to me. Do we need to add simple test cases? Or we can add in following patch.
Another question is do we need to add default values explicitly in submarine-site.xml? Maybe we can add that later too.

submarine-server/server-core/src/main/java/org/apache/submarine/server/JobManager.java
jiwq: Done.

jiwq:
Thanks @tangzhankun reviewing. The nits can be addressed in future, we should ensure the submarine server works well ASAP.

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfiguration.java
```diff
@@ -484,7 +523,10 @@ public void updateConfiguration(String name, String value) {
         "workbench.websocket.max.text.message.size", "1024000"),
     WORKBENCH_WEB_WAR("workbench.web.war", "submarine-workbench/workbench-web/dist"),
     SUBMARINE_RUNTIME_CLASS("submarine.runtime.class",
-        "org.apache.submarine.server.submitter.yarn.YarnRuntimeFactory");
+        "org.apache.submarine.server.submitter.yarn.YarnRuntimeFactory"),
+    SUBMARINE_SUBMITTERS("submarine.submitters", ""),
+    SUBMARINE_SUBMITTERS_ENTRY("submarine.submitters.*.class", ""),
+    SUBMARINE_SUBMITTERS_CLASSPATH("submarine.submitters.*.classpath", "");
```
liuxunorg:
```
    SUBMARINE_SUBMITTERS_ENTRY("submarine.submitters.*.class", ""),
    SUBMARINE_SUBMITTERS_CLASSPATH("submarine.submitters.*.classpath", "");
```
How to configure these 2 variables？
All PR submissions are finished, do you add any more Introduction documents?

jiwq:
As we discussed offline, I will close this PR now, and reopen it until all feature finished. So that we can save more time to ensure the server running well ASAP.

PR #136, opened 2019-12-30
SUBMARINE-327. Fix unchecked call 'result<T>' when used JsonResponse.Builder build response

jiwq:
### What is this PR for?
Fix the check warning for JsonResponse, the two warning problems list as below: 
* Unchecked call to 'result(T)' as a member of raw type
* Raw use of parameterized class

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-327

### Questions:


liuxunorg:
@jiwq Thank you fix this issue.

PR #147, opened 2020-01-11
SUBMARINE-340. Add submarine database dockerfile

liuxunorg:
### What is this PR for?
In order to enable submarine to support deployment in k8s,
We need to enable each service of submarine to be deployed independently.
The problem that jira solves is the ability to install the submarine table structure and initialization data into mysql.
Allows users to run submarine's database service with only one command.
like., 
```
docker run -it -p 3306:3306 -d --name s1 -e MYSQL_ROOT_PASSWORD=password apache/submarine:database-0.3.0-SNAPSHOT
```

### What type of PR is it?
[Refactoring]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/SUBMARINE-340

* https://travis-ci.org/liuxunorg/submarine/builds/635299028

### Questions:

### DIFF
--- a/dev-support/docker-images/database/Dockerfile
+++ b/dev-support/docker-images/database/Dockerfile
@@ -1,0 +1,26 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


pingsutw:
@liuxunorg Thanks for the patch, do we need a [docker-compose.yml](https://docs.docker.com/compose/) to let user easily launch different docker containers at the same time. Maybe file another Jira ticket.

liuxunorg:
> @liuxunorg Thanks for the patch, do we need a docker-compose.yml to let user easily launch different docker containers at the same time. Maybe file another Jira ticket.

Our main purpose of providing docker is to:
1. Allow users to quickly experience submarine through docker image.
2. Make submarine run in k8s through docker image.

That's enough. Because k8s is already the standard for container orchestration, we don't need to support `docker-compose`.

pingsutw:
Okay, I got it. Thanks @liuxunorg

PR #148, opened 2020-01-11
SUBMARINE-341. Add submarine server dockerfile

liuxunorg:
### What is this PR for?
In order to enable submarine to support deployment in k8s,
We need to enable each service of subarine to be deployed independently.
The problem that jira solves is that it can run the subarine server independently in docker.
Let users only need a docker command to run the submarine service.
like., 
```
docker run -it -d --link=submarine-database:submarine-database --name submarine apache/submarine:submarine-0.3.0-SNAPSHOT
```

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/SUBMARINE-341

* https://travis-ci.org/liuxunorg/submarine/builds/635308111

### Questions:

### DIFF
--- a/dev-support/docker-images/submarine/build.sh
+++ b/dev-support/docker-images/submarine/build.sh
@@ -16,7 +16,7 @@
 
 set -eo pipefail
 
-SUBMARINE_VERSION=0.3.0
+SUBMARINE_VERSION=0.3.0-SNAPSHOT
 SUBMARINE_IMAGE_NAME="apache/submarine:submarine-${SUBMARINE_VERSION}"
 
 if [ -L ${BASH_SOURCE-$0} ]; then
@@ -30,7 +30,7 @@ export SUBMARINE_HOME=${CURRENT_PATH}/../../..
 if [ ! -d "${SUBMARINE_HOME}/submarine-dist/target" ]; then
   mkdir "${SUBMARINE_HOME}/submarine-dist/target"
 fi
-submarine_dist_exists=$(find -L "${SUBMARINE_HOME}/submarine-dist/target" -name "submarine-dist-${submarine_v}*.tar.gz")
+submarine_dist_exists=$(find -L "${SUBMARINE_HOME}/submarine-dist/target" -name "submarine-dist-${SUBMARINE_VERSION}*.tar.gz")
 # Build source code if the package doesn't exist.
 if [[ -z "${submarine_dist_exists}" ]]; then
   # update tony code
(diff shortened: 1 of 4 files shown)


dev-support/docker-images/submarine/build.sh
```diff
@@ -0,0 +1,62 @@
...
+  mvn clean package -DskipTests
+fi
+
+mkdir -p "${CURRENT_PATH}/tmp"
+cp ${SUBMARINE_HOME}/submarine-dist/target/submarine-dist-${SUBMARINE_VERSION}*.tar.gz "${CURRENT_PATH}/tmp"
+
+# Replace the mysql jdbc.url in the submarine-site.xml file with the link name of the submarine container
+# `submarine-database` is submarine database container name
+cp ${SUBMARINE_HOME}/conf/submarine-site.xml "${CURRENT_PATH}/tmp/"
+sed -i ".bak" 's/127.0.0.1:3306/submarine-database:3306/g' "${CURRENT_PATH}/tmp/submarine-site.xml"
```
pingsutw: Thanks for the patch,@liuxunorg.
I got an error when running this line 
error message: 
``sed: -e expression #1, char 1: unknown command: `.'``

I change the command to ``sed -i.bak 's/127.0.0.1:3306/submarine-database:3306/g' "${CURRENT_PATH}/tmp/submarine-site.xml"``, and it works for me

dev-support/docker-images/submarine/build.sh
```diff
@@ -0,0 +1,62 @@
...
+else
+  PWD=$(dirname ${BASH_SOURCE-$0})
+fi
+export CURRENT_PATH=$(cd "${PWD}">/dev/null; pwd)
+export SUBMARINE_HOME=${CURRENT_PATH}/../../..
+
+if [ ! -d "${SUBMARINE_HOME}/submarine-dist/target" ]; then
+  mkdir "${SUBMARINE_HOME}/submarine-dist/target"
+fi
+submarine_dist_exists=$(find -L "${SUBMARINE_HOME}/submarine-dist/target" -name "submarine-dist-${submarine_v}*.tar.gz")
```
pingsutw: It's better to use `SUBMARINE_VERSION`

dev-support/docker-images/submarine/build.sh
liuxunorg: Done.

dev-support/docker-images/submarine/build.sh
liuxunorg: Done.

PR #161, opened 2020-01-21
SUBMARINE-356. ParameterProto should support getOptionValue method

yuanzac:
### What is this PR for?
To enable yaml configuration, the method of getOptionValue in ParameterHolder is used to keep job configuration.  ParameterProto should support this method as well.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-356

### How should this be tested?
https://travis-ci.org/yuanzac/hadoop-submarine/builds/639866700?utm_source=github_status&utm_medium=notification

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #179, opened 2020-02-10
SUBMARINE-377. [SDK] Submit a job by pysubamrine

pingsutw:
### What is this PR for?
Use python SDK to submit a submarine job.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-377

https://travis-ci.org/pingsutw/hadoop-submarine/builds/648336769

### Questions:

### DIFF
--- a/submarine-sdk/pysubmarine/tests/client/test_client.py
+++ b/submarine-sdk/pysubmarine/tests/client/test_client.py
@@ -1,0 +1,48 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from submarine.job import SubmarineJOBClient
+import mock
+import pytest
+import json
+
+
+@pytest.fixture(scope="function")
+def output_json_filepath():
+    dummy = {'a': 200, 'b': 2, 'c': 3}
+    path = '/tmp/data.json'
+    with open(path, 'w') as f:
+        json.dump(dummy, f)
+    return path
+
+
+@mock.patch('submarine.job.submarine_job_client.http_request')
+class TestSubmarineJobClient:
+    def test_submit_job(self, mock_http_request, output_json_filepath):
+        client = SubmarineJOBClient('submarine', 8080)
+        mock_http_request.return_value = {'jobId': 'job_1582524742595_0040',
+                                          'name': 'submarine', 'identifier': 'test'}
+        response = client.submit_job(output_json_filepath)
+
+        with open(output_json_filepath) as json_file:
+            json_body = json.load(json_file)
+
+        mock_http_request.assert_called_with('http://submarine:8080',
+                                             json_body=json_body,
+                                             endpoint='/api/v1/jobs', method='POST')
+
+        assert response['jobId'] == 'job_1582524742595_0040'
+        assert response['name'] == 'submarine'
+        assert response['identifier'] == 'test'
--- a/submarine-sdk/pysubmarine/example/client/README.md
+++ b/submarine-sdk/pysubmarine/example/client/README.md
@@ -11,7 +11,7 @@
 -->
 
 # Submarine Client API Example
-Tis example shows how you could use Submarine Client API to 
+This example shows how you could use Submarine Client API to 
 manage submarine tasks
 
 ## Prerequisites
--- a/submarine-sdk/pysubmarine/example/client/README.md
+++ b/submarine-sdk/pysubmarine/example/client/README.md
@@ -1,0 +1,32 @@
+<!---
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+   http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


submarine-sdk/pysubmarine/tests/client/test_client.py
```diff
@@ -0,0 +1,161 @@
...
+            application_id = 'application-123456789'
+
+        return mocked_response()
+
+    self.SubmitJob = mocked_submit
+
+
+# Submarine client parameters
+host = 'submarine'
+port = 8980
```
liuxunorg: `8980`, This is submarine rpc server port?

submarine-sdk/pysubmarine/tests/client/test_client.py
```diff
@@ -0,0 +1,161 @@
...
+keytab = 'keytabPath'
+principal = 'testPrincipal'
+distribute_keytab = True
+queue = 'submarine'
+client = Client(host=host, port=port, security_disabled=security_disabled, keytab=keytab,
+                principal=principal, distribute_keytab=distribute_keytab, queue=queue)
+
+# Submarine job parameters
+name = 'test_rpc'
+docker_image = 'apache/submarine:latest'
```
liuxunorg: Because submarine has many types of mirrors, such as: `submairne database`, `submarine server`, `submarine-operator`, now all use naming rules such `as apache/submarine:server-VERSION`.

If you need to use a `submarine-server` image here, you need to change it to `apache/submarine:server-latest`.

Regarding the command of mirroring, we may need to discuss it in the documentation or community meetings.

pingsutw: This is just an arbitrary string of docker image.
Actually, we won't run this docker images in the submarine cluster. 
I used mock_submit, so we won't pass these parameters to RPC server

submarine-sdk/pysubmarine/tests/client/test_client.py
pingsutw: @liuxunorg Thanks for the review.
Yes, this is default submarine rpc server port

submarine-sdk/pysubmarine/submarine/proto/SubmarineServerProtocol_pb2.py
```diff
@@ -0,0 +1,1075 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
yuanzac: Maybe we should not add the generated files in the git version management. For java, the protocolbuf generated files are located under target folder. So they are not managed by git. I'm not sure about how to work out for python.

submarine-server/server-submitter/submitter-yarn/src/main/java/org/apache/submarine/server/submitter/yarn/YarnUtils.java
```diff
@@ -46,129 +44,61 @@
...
-    tonyConf.setStrings(
-        TonyConfigurationKeys.getInstancesKey(Constants.WORKER_JOB_NAME),
-            parameters.getOptionValue(CliConstants.N_WORKERS));
-    if (parameters.getOptionValue(CliConstants.N_PS) != null) {
-      tonyConf.setStrings(
-              TonyConfigurationKeys.getInstancesKey(Constants.PS_JOB_NAME),
-              parameters.getOptionValue(CliConstants.N_PS));
-    }
-    // Resources for PS & Worker
-    if (parameters.getOptionValue(CliConstants.PS_RES) != null) {
```
yuanzac: The method of parameters.getOptionValue works for both cli arguments and yaml configuration.
PR #161 supports this method parameters.getOptionValue for grpc. Maybe we don't need to modify it.

liuxunorg:
Can this feature add e2e tests to `submarine-test/e2e`?

submarine-server/server-submitter/submitter-yarn/src/main/java/org/apache/submarine/server/submitter/yarn/YarnUtils.java
pingsutw: @yuanzac Thanks for the review.
> The method of parameters.getOptionValue works for both cli arguments and yaml configuration.

In YarnUtils `parameters.getOptionValue` will parse cli and yaml, and the write correspond conf to tony-xml, but when use pysubmarine client we won't pass `CommandLineProto` or yaml conf to rpc server. We pass `TensorFlowRunJobParameterProto` or `PyTorchRunJobParameterProto` , so that if we use `parameters.getOptionValue` , it can't find any parameter.

submarine-sdk/pysubmarine/submarine/proto/SubmarineServerProtocol_pb2.py
pingsutw: Let me check other python projects how to handle proto files.

pingsutw:
Thanks @liuxunorg for the suggestion. Could I file anthoer JIRA ticket for e2e tests

pingsutw:
Update patch. 
- remove pb2*.py that generated from protobuf
- update readme

submarine-server/server-submitter/submitter-yarn/src/main/java/org/apache/submarine/server/submitter/yarn/YarnUtils.java
yuanzac: Got you~
With this modification, does it still work when yaml is used to submit a job?
I just want to make sure that function is not affected.

pingsutw: Yes, I just try to submit job with YAML. it works well.

yuanzac: @pingsutw,
Thanks. you mean for both java client and python client, a job can be submitted with YAML, right?

pingsutw:
Update patch.

- A [README](https://github.com/pingsutw/hadoop-submarine/tree/SUBMARINE-377/submarine-sdk/pysubmarine/example/client) describes how to use sdk to submit a job
- Use REST to submit a job to submarine server
- Unit tests for the client (e2e test will add in next pr)

pingsutw:
@liuxunorg @yuanzac, Could you help me review the new patch

PR #181, opened 2020-02-10
SUBMARINE-384. Add EditorConfig metafile to reduce growth of the indentation mess

jiwq:
### What is this PR for?
To reduce growth of the indentation mess

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-384

### Questions:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,27 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# tab indentation (no size specified)
+[{Makefile,go.mod,go.sum,*.go}]
+indent_style = tab
+
+# 2 space indentation
+[*.{java,sh,json,xml}]
+indent_style = space
+indent_size = 2
+
+# 4 space indentation
+[*.{py,R,js}]
+indent_style = space
+indent_size = 4
+
+[*.java]
+max_line_length = 100
--- a/pom.xml
+++ b/pom.xml
@@ -466,6 +466,7 @@
             <exclude>.github/*</exclude>
             <exclude>.gitignore</exclude>
             <exclude>.gitmodules</exclude>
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,25 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# newline ending every file; no trailing space
```
yuanzac: If the comment is no trailing space. trim_trailing_whitespace should be true

.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+[{Makefile,go.mod,go.sum,*.go}]
+indent_style = tab
+
+# 2 space indentation
+[*.{java,json,sh}]
+indent_style = space
+indent_size = 2
+
+# 4 space indentation
+[*.{py,R,xml,js}]
```
yuanzac: Indentation size of pom.xml is 2

pingsutw:
There is an EditConfig file in `submarine-workbench/workbench-web/.editorconfig`, Do we need to combine to this config (`./submarine/.editorconfig`)

.editorconfig
jiwq: Done.

jiwq:
> With this EditorConfig file, what settings need to be set in the development tool (e.g., IDEA)?

High-frequency IDEs(IDEA/PyCharm/WebStorm/VisualStudio and so on) come bundled with native support for EditorConfig. So the user do not need extra work. More info see [here](https://editorconfig.org/).

.editorconfig
jiwq: Done.

jiwq:
> There is an EditConfig file in `submarine-workbench/workbench-web/.editorconfig`, Do we need to combine to this config (`./submarine/.editorconfig`)

@pingsutw Thanks for your focus, I think it's not conflicted. On the contrary, the subproject should better own itself EditorConfig. Any thoughts?

pingsutw:
Okay. If it's not conflicted, I thought we don't need to change it.
Thanks @jiwq. LGTM

PR #187, opened 2020-02-19
SUBMARINE-390. Add yarn submitter with classloader

yuanzac:
### What is this PR for?
Submitter dependencies are excluded from submarine server. Submarine rpc server should initialize yarn submitter through a classloader. 

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-390

https://travis-ci.org/yuanzac/hadoop-submarine/builds/652407783?utm_source=github_status&utm_medium=notification

### Questions:

### DIFF
--- a/bin/common.sh
+++ b/bin/common.sh
@@ -88,7 +88,7 @@ function download_mysql_jdbc_jar(){
   echo "Mysql jdbc jar is downloaded and put in the path of submarine/lib."
 }
 
-JAVA_OPTS+=" ${SUBMARINE_SERVER_JAVA_OPTS} -Dfile.encoding=UTF-8 ${SUBMARINE_SERVER_MEM}"
+JAVA_OPTS+=" -Dfile.encoding=UTF-8"
 JAVA_OPTS+=" -Dlog4j.configuration=file://${SUBMARINE_CONF_DIR}/log4j.properties"
 export JAVA_OPTS
 
--- a/conf/submarine-site.xml
+++ b/conf/submarine-site.xml
@@ -144,7 +144,7 @@
   </property>
 
   <property>
-    <name>submarine.server.remote.execution.enabled</name>
+    <name>submarine.server.rpc.enabled</name>
     <value>false</value>
     <description>Run jobs using rpc server.</description>
   </property>
--- a/conf/submarine-site.xml
+++ b/conf/submarine-site.xml
@@ -143,6 +143,12 @@
     <description>RuntimeFactory for Submarine jobs</description>
   </property>
 
+  <property>
+    <name>submarine.server.remote.execution.enabled</name>
+    <value>false</value>
+    <description>Run jobs using rpc server.</description>
+  </property>
+
   <!-- Submarine Submitters Configuration  -->
   <property>
     <name>submarine.submitters</name>
--- a/conf/submarine-site.xml
+++ b/conf/submarine-site.xml
@@ -149,6 +149,12 @@
     <description>Run jobs using rpc server.</description>
   </property>
 
+  <property>
+    <name>submarine.server.rpc.port</name>
+    <value>8980</value>
+    <description>Rpc server port</description>
+  </property>
+
   <!-- Submarine Submitters Configuration  -->
   <property>
     <name>submarine.submitters</name>
--- a/submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
+++ b/submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
@@ -44,10 +44,10 @@ public class SubmarineConfVars {
     SUBMARINE_SERVER_SSL_TRUSTSTORE_TYPE("submarine.server.ssl.truststore.type", null),
     SUBMARINE_SERVER_SSL_TRUSTSTORE_PASSWORD("submarine.server.ssl.truststore.password", null),
     SUBMARINE_CLUSTER_ADDR("submarine.cluster.addr", ""),
-    SUBMARINE_SERVER_REMOTE_EXECUTION_ENABLED(
-        "submarine.server.remote.execution.enabled", false),
-    SUBMARINE_SERVER_REMOTE_EXECUTION_PORT(
-        "submarine.server.remote.execution.port", 8980),
+    SUBMARINE_SERVER_RPC_ENABLED(
+        "submarine.server.rpc.enabled", false),
+    SUBMARINE_SERVER_RPC_PORT(
+        "submarine.server.rpc.port", 8980),
     CLUSTER_HEARTBEAT_INTERVAL("cluster.heartbeat.interval", 3000),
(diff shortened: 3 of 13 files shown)


bin/common.sh
```diff
@@ -88,7 +88,7 @@ function download_mysql_jdbc_jar(){
   echo "Mysql jdbc jar is downloaded and put in the path of submarine/lib."
 }
 
-JAVA_OPTS+=" ${SUBMARINE_SERVER_JAVA_OPTS} -Dfile.encoding=UTF-8 ${SUBMARINE_SERVER_MEM}"
+JAVA_OPTS+=" -Dfile.encoding=UTF-8"
```
liuxunorg: It's because `${SUBMARINE_SERVER_JAVA_OPTS}` and `${SUBMARINE_SERVER_MEM}"` are duplicated, so they need to be deleted, right?

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
```diff
@@ -109,11 +116,38 @@ private static ClientContext getClientContext(RpcContext rpcContext) {
     ClientContext clientContext = new ClientContext();
     clientContext.setYarnConfig(conf);
     mergeSubmarineConfiguration(clientContext.getSubmarineConfig(), rpcContext);
+    ClassLoader classLoader = new URLClassLoader(constructUrlsFromClasspath("../lib/submitter"));
```
liuxunorg: The `../lib/submitter` path will store `submitter-yarn` and `submitter-k8s` separately,
If we load all jars in the `../lib/submitter` path at the same time, there may be conflicts.
I think the RpcServer only needs to load the jar in the `submitter-yarn` or `submitter-k8s` path.
What do you think?

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
```diff
@@ -109,11 +116,38 @@ private static ClientContext getClientContext(RpcContext rpcContext) {
...
     RuntimeFactory runtimeFactory = RuntimeFactory.getRuntimeFactory(
-        clientContext);
+        clientContext, classLoader);
     clientContext.setRuntimeFactory(runtimeFactory);
     return clientContext;
   }
+  private static URL[] constructUrlsFromClasspath(String classpath) {
+    List<URL> urls = new ArrayList<>();
+    for (String path : classpath.split(File.pathSeparator)) {
+      if (path.endsWith("/*")) {
```
liuxunorg: Maybe no recursive loading jar?

liuxunorg:
@yuanzac 
Can you add `SUBMARINE-390.` to PR title?

bin/common.sh
yuanzac: Because they are added in submarine-daemon.sh and submarine.sh. So we need to delete them to avoid deplications.

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
yuanzac: It is not a recurisive loading jar. Only the jars in the folder will be loaded. The jars in the subfolder will not.

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
yuanzac: Yeah, agree with you. The code just adds jars in the submitter, not including jars in the submitter/submitter-k8s

yuanzac:
@jiwq,
Yeah, finally we will replace this logic with JobManager. But if we do this. we need to refactor the module structures to avoid dependency circles. After syncing with @liuxunorg and @tangzhankun last week, we decided to focus on HTTP API related development. For RPC part, we just make sure it works for what we have. When JobManager is done, it'll be used.

liuxunorg:
Yes, we prefer to provide external REST interfaces. The PRC interface will be perfected later.

conf/submarine-site.xml
```diff
@@ -143,6 +143,12 @@
     <description>RuntimeFactory for Submarine jobs</description>
   </property>
 
+  <property>
+    <name>submarine.server.remote.execution.enabled</name>
```
liuxunorg: Amended to `submarine.server.prc.enabled`, will it be clearer?

submarine-server/server-rpc/src/main/java/org/apache/submarine/server/rpc/SubmarineRpcServer.java
```diff
@@ -109,11 +118,45 @@ private static ClientContext getClientContext(RpcContext rpcContext) {
     ClientContext clientContext = new ClientContext();
     clientContext.setYarnConfig(conf);
     mergeSubmarineConfiguration(clientContext.getSubmarineConfig(), rpcContext);
+    String runtimeclass =
```
liuxunorg: Change `runtimeclass` to `runtimeClass`?

yuanzac: Done.

conf/submarine-site.xml
yuanzac: Done

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
```diff
@@ -44,10 +44,10 @@
     SUBMARINE_SERVER_SSL_TRUSTSTORE_TYPE("submarine.server.ssl.truststore.type", null),
     SUBMARINE_SERVER_SSL_TRUSTSTORE_PASSWORD("submarine.server.ssl.truststore.password", null),
     SUBMARINE_CLUSTER_ADDR("submarine.cluster.addr", ""),
-    SUBMARINE_SERVER_REMOTE_EXECUTION_ENABLED(
-        "submarine.server.remote.execution.enabled", false),
-    SUBMARINE_SERVER_REMOTE_EXECUTION_PORT(
-        "submarine.server.remote.execution.port", 8980),
+    SUBMARINE_SERVER_RPC_ENABLED(
+        "submarine.server.rpc.enabled", false),
+    SUBMARINE_SERVER_RPC_PORT(
```
liuxunorg: @yuanzac I think need add config entity `submarine.server.rpc.port` to `conf/submarine-site.xml` and `conf/submarine-site.xml.template`

like.
```
<property>
    <name>submarine.server.rpc.port</name>
    <value>8980</value>
    <description>Run jobs using rpc server pord.</description>
  </property>
```

yuanzac: Done.

PR #190, opened 2020-02-27
SUBMARINE-394. Change submit job response status in JobManagerRestApi

pingsutw:
### What is this PR for?
If job submit successfully we should return `OK`(200) instead of `ACCEPTED`(202) 
If job submit fail, success should be false

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-394

https://travis-ci.org/pingsutw/hadoop-submarine/builds/655167779

### Questions:


yuanzac:
Thanks @pingsutw and @liuxunorg. I'll merge it soon.

PR #211, opened 2020-03-09
SUBMARINE-411. Basic SQL Standard Authorization ACL Management for Spark

yaooqinn:
### What is this PR for?

This pr aims to provide submarine spark jobs with ACL Management abilities.
In this pr, we create
1) a plugin that can talk to Apache Ranger Admin directly for privilege policies
2) a Spark SQL optimizer rule that can be injected to spark jobs with the flag - spark.sql.extensions
3) the privilege check happens in the Spark SQL Optimizer meets a resource e.g. a database, table, etc. If the user has the privilege for the resource he or she accesses, we return the original spark plan and let it continue, otherwise, an exception will be raised 

The previous work can be found, this pr bring the ACL only and will do others in follow-ups
 - https://github.com/yaooqinn/spark-authorizer
 - https://github.com/yaooqinn/spark-ranger  
### What type of PR is it?
feature

### Todos
* [ ] - Show Databases/Tables with filtered objects
* [ ] - Row-level filtering
* [ ] - Datamasking filtering
* [ ] - Configuration Restriction

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-411

Add Unit test and enable Travis jobs
### Questions:
* Does this needs documentation? Yes, will do after finish all todos


yaooqinn:
cc @liuxunorg @yuanzac, thanks for reviewing in advance.

yaooqinn:
thanks @liuxunorg

yaooqinn:
notice that submarine supports Hadoop 3.1/3.2, which are not supported by spark 2.3/2.4. so I excluded them in Travis Hadoop 3.x related job in just the test phase.

liuxunorg:
Because there are several PRs to submit after this feature, I will merge this PR as soon as possible.

PR #218, opened 2020-03-10
SUBMARINE-417. Support Row Filtering

yaooqinn:
### What is this PR for?
Add row-level filtering feature to spark-security, which provider administrators to set system-level filters to end-users.

e.g. if table `src` has a system-level filter on column `key` (key<=20200309) for user bob

when bob execute `select * from src`, we will apply the filter to the query,  then the final result will be `select * from src where key<=20200309`

### What type of PR is it?

SUBMARINE-417

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/SUBMARINE-417/

added unit tests

### Questions:


yaooqinn:
cc @liuxunorg @yuanzac, thanks for reviewing in advance

PR #219, opened 2020-03-11
SUBMARINE-418. Support Data Masking for spark-security

yaooqinn:
### What is this PR for?

add data masking for spark-security, which enable administrators to mask sensitive data per column

e.g. phone num 12345678901 can be masked only show last 4 - nnnnnnn8901

### What type of PR is it?
feature
### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/SUBMARINE-418

add unit tests

### Questions:
* Does the licenses files need update? /No


yaooqinn:
cc @liuxunorg

PR #220, opened 2020-03-11
SUBMARINE-420. Full TPCDS test coverage for spark security plugin

yaooqinn:
### What is this PR for?

As we rewrite spark SQL plans in the spark-security plugin, we don't want to break anything AS-IS

ADD TPCS full queries support to enhance it.

### What type of PR is it?

Improvement

### Todos
* [x] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/SUBMARINE-420

UT
### Questions:
* Does the licenses files need update?No
* Is there breaking changes for older versions?No


yaooqinn:
cc @liuxunorg

PR #222, opened 2020-03-11
SUBMARINE-424. Enable scala style check in github actions

yaooqinn:
### What is this PR for?

add some scala style check rule from http://www.scalastyle.org/rules-1.0.0.html
enable it in the GitHub workflow
### What type of PR is it?

feature

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/SUBMARINE-424

pass github actions

### Questions:


yaooqinn:
cc @liuxunorg @yuanzac

PR #224, opened 2020-03-12
SUBMARINE-425. Optimize java style check in github actions

yaooqinn:
### What is this PR for?

the current java style check does not show the root cause for us.

### What type of PR is it?
Improvement
### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/SUBMARINE-425

pass GitHub action

### Questions:


yaooqinn:
cc @yuanzac @liuxunorg thanks guys

PR #226, opened 2020-03-12
SUBMARINE-84. Set random port in  TF_CONFIG

pingsutw:
### What is this PR for?
When there's no network virtualization solution like Calico deployed, the hard-coded 8000 port conflicts may cause worker/ps container gRPC server fails to start if they were allocated to one host and use host network.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-84

https://travis-ci.org/github/pingsutw/hadoop-submarine/builds/661489560
https://github.com/pingsutw/hadoop-submarine/actions/runs/54296809

### Questions:

### DIFF
--- a/submarine-server/server-submitter/submitter-yarnservice/src/main/java/org/apache/submarine/server/submitter/yarnservice/tensorflow/TensorFlowConfigEnvGenerator.java
+++ b/submarine-server/server-submitter/submitter-yarnservice/src/main/java/org/apache/submarine/server/submitter/yarnservice/tensorflow/TensorFlowConfigEnvGenerator.java
@@ -19,15 +19,18 @@
 
 package org.apache.submarine.server.submitter.yarnservice.tensorflow;
 
+import java.io.IOException;
+import org.apache.hadoop.net.ServerSocketUtil;
 import org.apache.submarine.commons.runtime.conf.Envs;
 import org.apache.submarine.server.submitter.yarnservice.YarnServiceUtils;
 
 public class TensorFlowConfigEnvGenerator {
 
   public static String getTFConfigEnv(String componentName, int nWorkers,
-      int nPs, String serviceName, String userName, String domain) {
+      int nPs, String serviceName, String userName, String domain) throws IOException {
     String commonEndpointSuffix = YarnServiceUtils
-        .getDNSNameCommonSuffix(serviceName, userName, domain, 8000);
+        .getDNSNameCommonSuffix(serviceName, userName, domain,
+          ServerSocketUtil.getPort(8000, 100));
 
     TFConfigEnv tfConfigEnv =
         new TFConfigEnv(nWorkers, nPs, componentName, commonEndpointSuffix);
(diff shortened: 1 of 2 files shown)


liuxunorg:
@pingsutw PR title is too long.

pingsutw:
@liuxunorg Thanks for the reminder. update it.

submarine-server/server-submitter/submitter-yarnservice/src/main/java/org/apache/submarine/server/submitter/yarnservice/tensorflow/TensorFlowConfigEnvGenerator.java
```diff
@@ -19,15 +19,18 @@
...
 import org.apache.submarine.server.submitter.yarnservice.YarnServiceUtils;
 
 public class TensorFlowConfigEnvGenerator {
 
   public static String getTFConfigEnv(String componentName, int nWorkers,
-      int nPs, String serviceName, String userName, String domain) {
+      int nPs, String serviceName, String userName, String domain) throws IOException {
     String commonEndpointSuffix = YarnServiceUtils
-        .getDNSNameCommonSuffix(serviceName, userName, domain, 8000);
+        .getDNSNameCommonSuffix(serviceName, userName, domain,
```
yuanzac: From my understanding, tensorflow config is generated before NM starts TF job worker containers. So the random port still may conflict with the ports on NM node. If the port is generated by NM, then the port conflictions would be resolved.

liuxunorg: Can we limit the port range to avoid this problem?

pingsutw: Thanks @yuanzac and @LinhaoZhu for the review.
make sense. I'm going to update it.

pingsutw: https://github.com/apache/submarine/blob/ebad135c1e44bc75f76c41fbaaeece87ed65d144/submarine-server/server-submitter/submitter-yarnservice/src/main/java/org/apache/submarine/server/submitter/yarnservice/YarnServiceJobSubmitter.java#L103-L114
@yuanzac, @liuxunorg 

since we use Yarn Service to submit the job, we couldn't control NM behavior.
so that we couldn't generate port in the NM node.
Please correct me if I'm wrong.

Maybe we could just close this pr since yarn service runtime is deprecated.

pingsutw: @tangzhankun Thanks for the reply.
Let me close it.

PR #227, opened 2020-03-12
SUBMARINE-428. Add job table in db to store job status.

yuanzac:
### What is this PR for?
We have to store job status in db. So that we don't need to get job status by k8s/yarn API every time.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-428

https://travis-ci.org/github/yuanzac/hadoop-submarine/builds/661526991?utm_source=github_status&utm_medium=notification

### Questions:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,4 +26,4 @@ indent_size = 4
 # Don't enforce indent size for java. maven check-style will handle other styles
 [*.java]
 indent_style = space
-max_line_length = 100
+max_line_length = 110
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/workbench/database/entity/Job.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/workbench/database/entity/Job.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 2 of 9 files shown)


liuxunorg:
@yuanzac Travis failure.

.editorconfig
```diff
@@ -14,7 +14,7 @@ trim_trailing_whitespace = true
 indent_style = tab
 
 # 2 space indentation
-[*.{java,sh,json,xml}]
+[*.{sh,json,xml}]
```
liuxunorg: Why deleted `java`? 
`java` uses 2 spaces for indentation.

submarine-server/server-core/src/main/java/org/apache/submarine/server/workbench/database/entity/Job.java
```diff
@@ -0,0 +1,93 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.submarine.server.workbench.database.entity;
+
+public class Job extends BaseEntity {
+
+  // (Todo) Add dict configuration.
```
liuxunorg:
```
// (Todo) Add dict configuration.
```
Is this an invalid comment, or will this feature need to be added later?
If you need to add this function in the future, you need to change it to `TODO (you name) Add dict configuration.`

.editorconfig
yuanzac: When a new line is added in an expression or a method defination, we need to use 4 spaces for indentation. But this configuration will use 2 spaces. This will lead an error for style checking.

submarine-server/server-core/src/main/java/org/apache/submarine/server/workbench/database/entity/Job.java
yuanzac: Thanks @liuxunorg, Done.

tangzhankun:
+1. Thanks @yuanzac . One minor question, can the jobService used by the submarine REST server directly?

PR #228, opened 2020-03-16
Fix conflicts of java indentation between .editorconfig and mvn plugin

tangzhankun:
### What is this PR for?
The ".editorconfig" enforced all the java indent size to 2 which will cause conflicts when maven check-style plugin does checking. We could remove this configuration from ".editorconfig".

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-429

Manually tested.
With this fix, the indent size in IDE like Intellij won't always be 2 and won't fail the mvn package.

### Questions:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ trim_trailing_whitespace = true
 indent_style = tab
 
 # 2 space indentation
-[*.{java,sh,json,xml}]
+[*.{sh,json,xml}]
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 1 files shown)


tangzhankun:
@jiwq @yuanzac Please review.

yuanzac:
LGTM.
The pr #227 delete "java" in line 17 as well.

tangzhankun:
Thanks for the review. @yuanzac , @liuxunorg . I'll merge it soon

PR #232, opened 2020-03-17
SUBMARINE-422. Fix refreshing page returns 404 error

liuxunorg:
### What is this PR for?
Because the workbench is developed using angular, the adjustment of angular WEB pages is completely controlled by the front end, so when you manually refresh a specific page in the browser, the browser will send the request for this page to the back-end service, but the back-end service only In response to API requests, it will cause the front end to display 404.

The solution is to find that not all API requests directly return the content of the index page,
so that the front end will automatically perform correct page routing processing.

### What type of PR is it?
[Bug Fix | Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/SUBMARINE-422

* https://travis-ci.org/github/liuxunorg/submarine/builds/663406363

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,12 @@ matrix:
         chrome: stable
         firefox: latest
       # Because submarine on k8s uses port 80, it needs to be set `-Durl=http://127.0.0.1`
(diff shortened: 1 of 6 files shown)


liuxunorg:

PR #242, opened 2020-03-21
SUBMARINE-448. Make Spark Security's RangerSparkPlugin static

yaooqinn:
### What is this PR for?

Now the RangerSparkPlugin is non-static and will be created and init times and times again, this should be an object, not a class in scala, to avoid high load policies decoding.

### What type of PR is it?
Refactoring
### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-448

existing tests

### Questions:


yaooqinn:
cc @liuxunorg

liuxunorg:
@yaooqinn PR title is too long. I modify.

PR #250, opened 2020-03-26
SUBMARINE-456. Fixed workbench-web Angular fails in travis

cchung100m:
### What is this PR for?
Hi @jojochuang @liuxunorg 

This PR is trying to fix the e2e angular testing job fails intermittently. The problem is that file `gecko-response.json` is downloaded from api.github.com so that if the jobs make too many requests at a given moment it fails.

I suggest that we can merge this PR and observe the CI jobs for the next few days or other commits before closing the Jira issue ticket.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-456

The PR is tested many times on my CI jobs and it builds successfully.

https://travis-ci.com/github/cchung100m/submarine/builds/155547853

螢幕快照 2020-03-25 下午11 00 28

### Questions:

### DIFF
--- a/submarine-workbench/workbench-web-ng/package.json
+++ b/submarine-workbench/workbench-web-ng/package.json
@@ -8,7 +8,7 @@
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
-    "webdriver": "webdriver-manager update"
+    "webdriver": "webdriver-manager update --ignore_ssl = true"
   },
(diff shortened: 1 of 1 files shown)


liuxunorg:
@cchung100m Thank you fixed this bug.

jojochuang:
Thanks for fixing it!

PR #249, opened 2020-03-25
SUBMARINE-455. Support CRUD job in submarine-server REST

jiwq:
### What is this PR for?
Nowadays the submarine-server only support create the job by REST. We should implement the other operations.

- GET/PATCH/DELETE job by REST
- Fill more info for Job object, such as the job status(accepted/created/running/succeeded) with time and the spec etc.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-455

### How should this be tested?
https://travis-ci.com/github/jiwq/submarine/builds/155090047

### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #253, opened 2020-03-29
SUBMARINE-300. Add UT for SubmitterManager

jiwq:
### What is this PR for?
To ensure the SubmitterManager can load any custom JobSubmitter which is set in configuration file.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-300

https://travis-ci.com/github/jiwq/submarine/builds/156283502

### Questions:

### DIFF
--- a/submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfiguration.java
+++ b/submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfiguration.java
@@ -338,7 +338,7 @@ public class SubmarineConfiguration extends XMLConfiguration {
    * @return class name
    */
   public String getSubmitterClass(String name) {
-    return getStringValue(String.format(SubmarineConfVars.ConfVars.SUBMARINE_SUBMITTERS_CLASS.getVarName(), 
+    return getStringValue(String.format(SubmarineConfVars.ConfVars.SUBMARINE_SUBMITTERS_CLASS.getVarName(),
         name), "");
   }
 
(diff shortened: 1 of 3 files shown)


cchung100m:
Hi @jiwq 

It seems like the e2e angular testing error did not solved by PR #250 . I re-open the issue [456](https://issues.apache.org/jira/browse/SUBMARINE-456) for tracking this error. Please re-trigger the build process again, thank you.

liuxunorg:
We only need one configuration parameter, and the `submitter_type` is equal to `k8s` or `yarn` to meet our needs.

There are several problems with supporting several resource scheduling systems at the same time:
1. You need to specify equal to k8s or yarn in the type of the job, so that the job has no universality. For example: the yarn I use and the k8s you use, then you must modify my job before you can use it.
2. I don't think it is necessary for `submarine` to support yarn and k8s at the same time.
3. Supporting multiple scheduling systems at the same time will make the program complicated and difficult to understand.

In addition, the following two configuration parameter values can be fixed and do not require dynamic configuration, which increases user confusion.

```
    SUBMARINE_SUBMITTERS_CLASS("submarine.submitters.%s.class", ""),
    SUBMARINE_SUBMITTERS_CLASSPATH("submarine.submitters.%s.classpath", ""),
```

jiwq:
Thanks @liuxunorg for reviewing. Good catch and idea, maybe we can create new JIRA tracking the optimization. Any thoughts?

jiwq:
As we discussed offline with @liuxunorg , I will close the PR, because it duplicate with #249 .

PR #266, opened 2020-04-17
SUBMARINE-467. Simplify the submitter configuration

jiwq:
### What is this PR for?
At now the submarine server supports config more submitters, but it's not necessary in current stage. At the same time it also cause trouble to users. So we'd better simplify the configuration to keep only one submitter in server. Considering the project vision, we set the K8s submitter as the default.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-467

https://travis-ci.com/github/jiwq/submarine/builds/160762675

### Questions:


submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
```diff
@@ -30,44 +31,49 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(SubmitterManager.class);
 
-  private SubmarineConfiguration conf;
-  private ConcurrentMap<String, JobSubmitter> submitterMap = new ConcurrentHashMap<>();
+  private static final SubmitterManager INSTANCE = new SubmitterManager();
 
-  public SubmitterManager(SubmarineConfiguration conf) {
-    this.conf = conf;
-    loadSubmitters();
+  private final Map<String, JobSubmitterConfig> SUBMITTER_TABLE = new HashMap<>();
```
yuanzac, line 46: We'd better rename SUBMITTER_TABLE to something like submitter_class_config.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
```diff
@@ -30,44 +31,49 @@
...
+
+  private JobSubmitter submitter;
+
+  {
```
yuanzac, line 50: We'd better move the initialization logic to a constructor method.

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
```diff
@@ -68,10 +68,7 @@
     WORKBENCH_WEB_WAR("workbench.web.war", "submarine-workbench/workbench-web/dist"),
     SUBMARINE_RUNTIME_CLASS("submarine.runtime.class",
         "org.apache.submarine.server.submitter.yarn.YarnRuntimeFactory"),
-    SUBMARINE_SUBMITTERS("submarine.submitters", ""),
-    SUBMARINE_SUBMITTERS_CLASS("submarine.submitters.%s.class", ""),
-    SUBMARINE_SUBMITTERS_CLASSPATH("submarine.submitters.%s.classpath", ""),
-    SUBMARINE_K8S_KUBE_CONFIG("submarine.k8s.kube.config", "");
+    SUBMARINE_SUBMITTER("submarine.submitter", "");
```
yuanzac, line 71: How about we make the fault value of SUBMARINE_SUBMITTER to k8s.

liuxunorg: Please set the default value to k8s

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
liuxunorg: yes, The name SUBMITTER_TABLE is too strange.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
```diff
@@ -92,11 +98,13 @@ private void loadSubmitters() {
...
 
-  /**
-   * Get the specified submitter by submitter type
-   * @return submitter
-   */
-  public JobSubmitter getSubmitterByType(String type) {
-    return submitterMap.get(type);
+  private static class JobSubmitterConfig {
+    private final String clazzName;
+    private final String classpath;
```
liuxunorg, line 103: Please change `classpath` to `classPath`.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
```diff
@@ -92,11 +98,13 @@ private void loadSubmitters() {
...
+
+    JobSubmitterConfig(String clazzName, String classpath) {
+      this.clazzName = clazzName;
+      this.classpath = classpath;
```
liuxunorg, line 107: Please change `classpath` to `classPath`.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
jiwq: Good point, I had fix it.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
jiwq: Using the initializing instance members to split the initialize logic and this approach can be used to share a block of code between multiple constructors. So that it can avoid some mistakes.
More info see https://docs.oracle.com/javase/tutorial/java/javaOO/initial.html

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
jiwq: Done.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
jiwq: The [classpath](https://en.wikipedia.org/wiki/Classpath_(Java)) is a parameter in JVM, so I think we should keep this naming convention.

submarine-server/server-core/src/main/java/org/apache/submarine/server/SubmitterManager.java
jiwq: The [classpath](https://en.wikipedia.org/wiki/Classpath_(Java)) is a parameter in JVM, so I think we should keep this naming convention.

jiwq:
Thanks @yuanzac and @liuxunorg help to review. 
I read your opinions and answered each question. If I understand error, please correct me.

yuanzac:
@jiwq, thanks for the contribution. I've no more questions. 
@liuxunorg, @tangzhankun any comments?

PR #269, opened 2020-04-27
SUBMARINE-484. DCL Framework: Basic antlr4 grammar support with create role as an example

yaooqinn:
### What is this PR for?

Part of DCL(Data Control Language) Framework Support

1. basic antlr4 parser for DCL SQLs
2. create role syntax added as an example

### What type of PR is it?

Feature

### Todos
* [ ] - Task

### What is the Jira issue?
 https://issues.apache.org/jira/browse/SUBMARINE-484

new unit tests

### Questions:


yaooqinn:
cc @liuxunorg, now I am working on the DCL framework support for Spark SQL, this is the very first commit that brings basic grammar and `create role` syntax as a POC example. thanks for reviewing.

PR #299, opened 2020-05-25
SUBMARINE-507. Submarine Environment Management

manirajv06:
### What is this PR for?
https://issues.apache.org/jira/browse/SUBMARINE-507

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-507

### Questions:


tangzhankun:
@manirajv06 , Thanks for your contribution! The direction looks good to me!

submarine-server/server-core/src/main/java/org/apache/submarine/server/rest/EnvironmentRestApi.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+  
+  @PATCH
+  @Path("/{id}")
+  @Consumes({RestConstants.MEDIA_TYPE_YAML, MediaType.APPLICATION_JSON})
+  @Operation(summary = "Update the environment with job spec",
+          tags = {"environments"},
+          responses = {
+                  @ApiResponse(description = "successful operation", content = @Content(
+                          schema = @Schema(implementation = Environment.class))),
```
pingsutw, line 84:
```suggestion
                          schema = @Schema(implementation = JsonResponse.class))),
```

liuxunorg:
@manirajv06 Your PR title is too long and has been cut into two parts, need change to short.

liuxunorg:
@manirajv06 The title of PR should be consistent with that of JIRA.
I fixed it for you.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
```diff
@@ -0,0 +1,100 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.submarine.server.api.spec;
+
+import java.util.Map;
```
liuxunorg, line 22: Please delete this line.  Here a style check error is triggered.
```
Unused import - java.util.Map.
```

tangzhankun:
@manirajv06 Thanks for the contribution. It generally looks good to me. One question:
did you test the "condo" commands in init_container locally? I don't see any integration test in test-k8s module.

manirajv06:
@pingsutw Fixed 

> Thanks @manirajv06
> Overall LGTM, a few nits.
> 
> 1. ApiResonse class of swagger should be `JsonResponse.class`
> 2. The license is missing in some files

@pingsutw  Thanks. Taken care.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
manirajv06: Thanks for fixing the title. Fixed check style issues also.

manirajv06:
> @manirajv06 Thanks for the contribution. It generally looks good to me. One question:
> did you test the "condo" commands in init_container locally? I don't see any integration test in test-k8s module.

Yes, able to see init container process working as expected. Initial, there were some issues. Passing commands as part of init containers in appropriate way had fixed the problem. Through k8's pod logs, can see the dependencies getting installed with in environment and init container 's "Ready" flag also has been set to True. However, actual python training job didn't complete because of memory issues and final state of the master and worker pod has changed to OOMKilled.

tangzhankun:
@manirajv06 

> Yes, able to see init container process working as expected. Initial, there were some issues. Passing commands as part of init containers in appropriate way had fixed the problem. Through k8's pod logs, can see the dependencies getting installed with in environment and init container 's "Ready" flag also has been set to True. However, actual python training job didn't complete because of memory issues and final state of the master and worker pod has changed to OOMKilled.

Cool. Maybe we can use a "pip show json" to verify if the dependencies are installed?

BTW, the job API has been refactored to experiment API. could you please rebase this PR?

liuxunorg:
@manirajv06 
> However, actual python training job didn't complete because of memory issues and final state of the master and worker pod has changed to OOMKilled.

Is your test tested in `Test submarine test-k8s` in travis?
If there is a problem, please raise it in the next community meeting and we will solve it together.
In addition, `Test submarine test-k8s` does need to be optimized.

manirajv06:
@tangzhankun @wangdatan ExperimentAPI has introduced "environment" param as part of Job/Experiment spec. For example,

  "environment": {
    "image": "apache/submarine:pytorch-dist-mnist-1.0"
  }

I've also introduced "environment" param as part of this JIRA like given below:

"environment": "my-submarine-env"

We can do the following:

1. Either differentiate the value based on "Type" (either string or jsonobject) by retaining the same key "environment" and build out the next steps.
2. We can rename one key to "predefined-environment" or "named-environment" or "environment-template" etc. For example,

"environment-template": "my-submarine-env"

Thoughts?

tangzhankun:
@manirajv06 . Yeah. Although that refactor has an environment field,
it's just a beginning to split the docker image to your new environment spec. The final environment API/Spec should base on your PR.

To me, use the spec in this PR as below to set up a pre-defined environment is the step to register a good environment.
```
{
  "name": "my-submarine-env",
  "dockerImage" : "continuumio/anaconda3",
  "kernelSpec" : {
  	"name" : "team_default_python_3.7",
  	"channels" : ["defaults"],
  	"dependencies" : 
  		["_ipyw_jlab_nb_ext_conf=0.1.0=py37_0", 
  		"alabaster=0.7.12=py37_0",
  		"anaconda=2020.02=py37_0",
  		"anaconda-client=1.7.2=py37_0",
  		"anaconda-navigator=1.9.12=py37_0"]
  }
}
```

And then using it with a pre-defined environment name. 

```
"environment": "my-submarine-env",
"experimentMeta":{...},
"spec":{...}
```

Now the problem is whether we should still support using the below environment spec with only one docker image in it.
```
{
"environment": {
  "image": "apache/submarine:pytorch-dist-mnist-1.0"
}
"experimentation":{...},
"spec":{...}
}
```

My option is that we should still support this as an **anonymous environment**. It's used for quick testing of the environment. Once he's sure that this image is good, he should consider use the docker image + kernel spec to register the finalized environment. And then use it again and again.

So it doesn't change how to create an environment. But when using the environment spec:
It can be either
```
{
"environment": {
"image": "apache/submarine:pytorch-dist-mnist-1.0"
}
"experimentation":{...},
"spec":{...}
}
```
Or
```
{
"environment": {
"name": "my-submarine-env"
}
"experimentation":{...},
"spec":{...}
}
```
When both are using, it reports an error.

Thoughts? @jiwq , @wangdatan , @manirajv06 ?

manirajv06:
> My option is that we should still support this as an anonymous environment. It's used for quick testing of the environment. Once he's sure that this image is good, he should consider use the docker image + kernel spec to register the finalized environment. And then use it again and again.

Do you mean to say, users would be keep on installing python deps one by one / as and when they see the dependency warning messages after first round of testing? Then, uses the full list of deps as kernel spec once for all during final registration of environment?

In reality, users would be knowing the deps before in hand. Right? Then, Why?

However, I do see a point in anonymous environment as @wangdatan described earlier. It makes users life easier using environment on the fly rather than creating and using it provided environment spec is very simple and has only docker image. Do we have any idea on no. of use cases falls under this category? If not many, Given that we have API support for all environment CRUD operations, Can we make it simple with support only for "name" ? These API's support (and SDK support eventually) would be very handy for users. Thoughts?

tangzhankun:
@manirajv06 We're almost ready to ship 0.4.0 release. If we remove the "image" field after this PR and only support a "name" in the `environment`. It will break the API.
Another reason I want to keep it is that there might be the use case that the user doesn't want to register an environment beforehand.
He just wants to quickly try the application with one ready image which might be built by someone else. He even doesn't want what's installed in it. Once he runs an experiment with the image successfully, he might consider registering it without any conda spec.
If we don't support this, it seems not very flexible to me.
```
environment: {
image:...
}
```

BTW, do we support just register a docker image as an environment without kernel spec?

{
  "name": "my-submarine-env",
  "dockerImage" : "someone shared based image",
}

manirajv06:
Ok. To make the code implementation simpler and also to provide a well defined consistent spec construct, can we have the following in Experiment spec?

For embedded/anonymous environment:

"environment": {
"image": "apache/submarine:pytorch-dist-mnist-1.0"
}

For Named/Pre-defined environment:

environment: {
"name": "my-submarine-env",
}

instead of

"environment": "my-submarine-env"

Above environment spec construct is easy for implementation as we don't need to decide based on the type of value (string or json object) as well as intuitive for users. Since structure is same, we can have single EnvironmentSpec.java class (introduced in this JIRA) itself to extract the values from spec/json and can have a additional method some thing like 

  public boolean isEmbeddedEnvironment() {
    return (getImage() != null && getName() == null);
  }

to know whether environment is embedded or not.

In addition, ExperimentSpecParser can also the following checks:

    if (experimentSpec.getEnvironmentSpec().getName() != null
        && experimentSpec.getEnvironmentSpec().getImage() != null) {
      throw new InvalidSpecException("Invalid environment spec: Either "
          + "image '" + experimentSpec.getEnvironmentSpec().getImage() + "' or "
          + "name '" + experimentSpec.getEnvironmentSpec().getName()
          + "' is allowed.");
    }

Infact, with this construct, users could play around even with kernel spec also in embedded environment fashion itself during iterations and finally go for environment registration. For example, Experiment spec could have environment (if needed) in the following way:

{
"environment": {
"name": "my-submarine-env",
  "dockerImage" : "continuumio/anaconda3",
  "kernelSpec" : {
  	"name" : "team_default_python_3.7",
  	"channels" : ["defaults"],
  	"dependencies" : 
  		["_ipyw_jlab_nb_ext_conf=0.1.0=py37_0", 
  		"alabaster=0.7.12=py37_0",
  		"anaconda=2020.02=py37_0",
  		"anaconda-client=1.7.2=py37_0",
  		"anaconda-navigator=1.9.12=py37_0"]
  }
}
"experimentation":{...},
"spec":{...}
}

Once users has reached the final stage, above Experiment spec could be changed to:

{
"environment": {
"name": "my-submarine-env",
}
"experimentation":{...},
"spec":{...}
}

provided environment ""my-submarine-env" has been registered before in hand.

@tangzhankun @wangdatan  Thoughts?

tangzhankun:
@manirajv06 , Thanks for the update. I like the new environment spec in experiment spec. Using the same struct can simplify our code.

And regarding this, will it register the "my-submarine-env" implicitly?
```
{
"environment": {
"name": "my-submarine-env",
"dockerImage" : "continuumio/anaconda3",
"kernelSpec" : {
"name" : "team_default_python_3.7",
"channels" : ["defaults"],
"dependencies" :
["_ipyw_jlab_nb_ext_conf=0.1.0=py37_0",
"alabaster=0.7.12=py37_0",
"anaconda=2020.02=py37_0",
"anaconda-client=1.7.2=py37_0",
"anaconda-navigator=1.9.12=py37_0"]
}
}
"experimentation":{...},
"spec":{...}
}
```

manirajv06:
Recent commit contain changes based on our discussion. EnvironmentSpec handles both anonymous/embedded environment and Named/Pre-defined environment. Either one approach is allowed as of now in ExperimentSpec.

Have 1 general question:

As part of ExperimentAPI refactoring work, "image" in "LibrarySpec" (based on old job spec) has been moved to 

"environment": {
"image":....
}

For example, 

  "environment": {
    "image": "apache/submarine:pytorch-dist-mnist-1.0"
  },

By any chance, Does users need both "apache/submarine:pytorch-dist-mnist-1.0" (copied from exp spec "image" ) and "continuumio/anaconda3" (copied from environment spec "dockerImage". Please refer https://github.com/apache/submarine/pull/299/commits/3b10a9e23236f62994412f3bbb48ae0ae9cd52af#diff-14695911fd5696c8184159015e2fdf26 ) ?

> And regarding this, will it register the "my-submarine-env" implicitly?

We can do with extra flag "auto-save" or something like that either through spec or some other way to make it intuitive for users?

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/environment/Environment.java
```diff
@@ -0,0 +1,64 @@
...
+
+package org.apache.submarine.server.api.environment;
+
+import org.apache.submarine.server.api.spec.EnvironmentSpec;
+
+public class Environment {
+
+  /**
+   * ID of the environment
+   */
```
wangdatan, line 28: Why we need an ID here? is a unique name sufficient?

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
```diff
@@ -0,0 +1,117 @@
...
+ * @author mani
+ *
+ */
+public class EnvironmentSpec {
+
+  
+  /**
+   * Name of the environment
+   */
+  private String name;
```
wangdatan, line 35: We have name in both Environment and EnvironmentSpec, should we dedup it? (I suggest to leave it in the Environment only).

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
```diff
@@ -0,0 +1,117 @@
...
+  public void setDescription(String description) {
+    this.description = description;
+  }
+
+  /**
+   * @return the image
+   */
+  public String getImage() {
+    return image;
+  }
```
wangdatan, line 101: I guess this is for VM image, can we rename it to vmimage?

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
```diff
@@ -0,0 +1,226 @@
...
+  }
+  
+  /**
+   * Update environment
+   * @param name Name of the environment
+   * @param spec environment spec
+   * @return Environment environment
+   * @throws SubmarineRuntimeException the service error
+   */
+  public Environment updateEnvironment(String name, EnvironmentSpec spec)
```
wangdatan, line 110: It looks like updateEnvironment and createEnvironment have duplicated logic, can you try to dedup?

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
```diff
@@ -0,0 +1,226 @@
...
+  
+  /**
+   * Get Environment
+   * @param name Name of the environment
+   * @return Environment environment
+   * @throws SubmarineRuntimeException the service error
+   */
+  public Environment getEnvironment(String name)
+      throws SubmarineRuntimeException {
+    Environment environment = getEnvironmentDetails(name);
```
wangdatan, line 176: Is this always load from DB, is it better to maintain an in-mem cache?

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
```diff
@@ -158,6 +172,49 @@ private static V1PodTemplateSpec parseTemplateSpec(ExperimentTaskSpec taskSpec,
...
+        V1Container initContainer = new V1Container();
+        initContainer.setName(environment.getName());
+        initContainer.setImage(environmentSpec.getDockerImage());
+
+        if (environmentSpec.getKernelSpec().getDependencies().size() > 0) {
+          StringBuffer createCommand = new StringBuffer();
+          String condaEnvironmentName =
+              environmentSpec.getKernelSpec().getName();
+
+          createCommand.append("conda create -n " + condaEnvironmentName);
```
wangdatan, line 193: Oh yeah, I got the conda related implementation. So basically this added a init container for conda dependency. Do you get chance to test this? Is it actually work? 

Also, it is important to make a better debuggability, for example, this hardcoded a conda path to /opt/conda/envs/env/bin. And i'm not sure if write something to `~/.bashrc` is portable enough or not. I'd like to hear thoughts from @sunilgovind. 

Another thing is, this assumes conda is part of Docker image, what if the Conda is not there, and if that happens, how we can give useful information to users if conda is not installed. 

Also, I think we need to check minimum (or maybe maximum) version of conda.

wangdatan:
also, I didn't see APIs to save image, and examples to use Conda images from Python spec, can we add them? @manirajv06

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/environment/Environment.java
manirajv06: Name is sufficient. In general, Having auto increment identifier for table is a good practice to do any CRUD operations in future. Just used that in Environment object as well.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
manirajv06: Ok. Since name is part of EnvironmentSpec, kept in EnvironmentSpec class itself.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
manirajv06: Had vmimage in the very beginning. Then removed it after your comments mentioned on https://issues.apache.org/jira/browse/SUBMARINE-507?focusedCommentId=17116299&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17116299.

Now, it had come again because of recent commit - to handle below environment definition

environment: {
"image"...
}

of ExperimentSpec in EnvironmentSpec itself. In fact, raised a question about the same also. Please refer https://github.com/apache/submarine/pull/299#issuecomment-648614756.

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
manirajv06: updateEnvironment and createEnvironment calls update and insert method of db mapper class respectively. Hence the different methods. (or) Am I missing anything?

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
manirajv06: Taken care.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
manirajv06: > Do you get chance to test this? Is it actually work?

Yes. Please refer https://github.com/apache/submarine/pull/299#issuecomment-644230611 for more details.

> for example, this hardcoded a conda path to /opt/conda/envs/env/bin. And i'm not sure if write something to `~/.bashrc` is portable enough or not. I'd like to hear thoughts from @sunilgovind.

Had come across mutliple ways while working on it and finally used the approach described here - 
https://medium.com/@chadlagore/conda-environments-with-docker-82cdc9d25754

But, yes, we can improve further.

> Another thing is, this assumes conda is part of Docker image, what if the Conda is not there, and if that happens, how we can give useful information to users if conda is not installed.

Yes, even used the examples/test cases with "continuumio/anaconda3" as dockerImage. This question kind of answer my question mentioned in https://github.com/apache/submarine/pull/299#issuecomment-648614756 ? In this case, users will need to build a new image containing both the stuff?

Will think about doing conda existence check..

> Also, I think we need to check minimum (or maybe maximum) version of conda

Why? Can you explain? Do we need to inform users in case version is not as expected?

manirajv06:
> also, I didn't see APIs to save image, and examples to use Conda images from Python spec, can we add them? @manirajv06

manirajv06:
> > also, I didn't see APIs to save image, and examples to use Conda images from Python spec, can we add them? @manirajv06

manirajv06:
> I didn't see APIs to save image

Do you mean pushing to "apache/submarine" repository?

> examples to use Conda images from Python spec,

Do you mean executing some commands inside activated conda env to confirm its working?

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/environment/Environment.java
wangdatan: Make sense. However, I suggest using a string for ID which is more flexible and consistent with other best practices (like K8s). Incremental ID could cause clashing for HA cases.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
wangdatan: I see, then I suggest to rename it to "containerImage" to make it more specific. "Image" maybe too general.

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
wangdatan: There's only one line difference between the method, it's better to consolidate the two methods. Since it is not a trivial method, dedup can help us reduce maintenance cost, thus I suggested to do that.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
wangdatan: I'd like to leave detailed review of this part to @sunilgovind or @liuxunorg. I'm not a shell expert. 

> Why? Can you explain? Do we need to inform users in case version is not as expected?

The reason is to avoid unexpected incompatible changes across different versions of Conda. It will be fine if we can give a major (minor) version range as recommendation. Doing a check at the runtime can save people' time to troubleshoot these issues.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/environment/Environment.java
manirajv06: Taken care.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
manirajv06: Experiment Spec contains "image" and we are about to ship 0.4.0 release. So, renaming as "containerImage" cause unnecessary confusions in upcoming releases. If we want to rename, it is better to do for 0.4.0 release itself. Thoughts?

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
manirajv06: Taken care.

manirajv06:
> > I didn't see APIs to save image
> 
> Do you mean pushing to "apache/submarine" repository?
> 
> > examples to use Conda images from Python spec,
> 
> Do you mean executing some commands inside activated conda env to confirm its working?

Can you please check this as well?

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/Environment.java
```diff
@@ -20,18 +20,43 @@
...
+   * ID of the environment
+   */
+  private int environmentId;
+  
   /**
-   * Get the image for the experiment. If the {@link ExperimentTaskSpec#getImage()} not
-   * specified the image replaced with it.
-   * @return image
+   * Name of the environment
    */
```
wangdatan, line 31: It looks like the name is still here.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
wangdatan: @manirajv06 , sorry I missed your last comment, we now have both dockerImage and image, I think we should only leave dockerImage. 

From 0.4.0 to 0.5.0 it is fine to break compatibility, since 0.4.0 we explicitly called out that it is an alpha release. Other experiment API will be unavoidably updated too (in an incompatible way).

submarine-server/server-core/src/main/java/org/apache/submarine/server/environment/EnvironmentManager.java
wangdatan: Thanks.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
wangdatan: @manirajv06 can you check my comment above?

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/Environment.java
manirajv06: Taken care. Class itself is redundant, so removed.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
manirajv06: Slightly changed the definition of "environment" in Experiment spec. For example, Below spec is valid.

environment: {
"image":"..",
"name":".."
}

If "image" is available, set it as k8 container image. If "name" is available, fetch the environment details using name from DB/Cache and use the details for K8 init container process. Hope this make the whole thing intuitive and simpler. Thoughts? @tangzhankun @wangdatan

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
manirajv06: Taken care. Conda version checks are done. It should fall between min and max version (configurable).

manirajv06:
> > > I didn't see APIs to save image
> > 
> > 
> > Do you mean pushing to "apache/submarine" repository?
> > > examples to use Conda images from Python spec,
> > 
> > 
> > Do you mean executing some commands inside activated conda env to confirm its working?
> 
> Can you please check this as well?

Can you please check the questions above?

liuxunorg:
@manirajv06 @wangdatan Can this PR be merged?

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/AbstractUniqueIdGenerator.java
```diff
@@ -0,0 +1,93 @@
+package org.apache.submarine.commons.utils;
```
jiwq, line 1: Missing the header license.

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
```diff
@@ -68,7 +68,9 @@
     WORKBENCH_WEB_WAR("workbench.web.war", "submarine-workbench/workbench-web/dist"),
     SUBMARINE_RUNTIME_CLASS("submarine.runtime.class",
         "org.apache.submarine.server.submitter.yarn.YarnRuntimeFactory"),
-    SUBMARINE_SUBMITTER("submarine.submitter", "k8s");
+    SUBMARINE_SUBMITTER("submarine.submitter", "k8s"),
+   ENVIRONMENT_CONDA_MIN_VERSION("environment.conda.min.version", "4.0.1"),
+   ENVIRONMENT_CONDA_MAX_VERSION("environment.conda.max.version", "4.10.10");
```
jiwq, line 73:
```suggestion
    ENVIRONMENT_CONDA_MAX_VERSION("environment.conda.max.version", "4.10.10");
```

submarine-commons/commons-utils/src/main/java/org/apache/submarine/commons/utils/SubmarineConfVars.java
```diff
@@ -68,7 +68,9 @@
     WORKBENCH_WEB_WAR("workbench.web.war", "submarine-workbench/workbench-web/dist"),
     SUBMARINE_RUNTIME_CLASS("submarine.runtime.class",
         "org.apache.submarine.server.submitter.yarn.YarnRuntimeFactory"),
-    SUBMARINE_SUBMITTER("submarine.submitter", "k8s");
+    SUBMARINE_SUBMITTER("submarine.submitter", "k8s"),
+   ENVIRONMENT_CONDA_MIN_VERSION("environment.conda.min.version", "4.0.1"),
```
jiwq, line 72:
```suggestion
    ENVIRONMENT_CONDA_MIN_VERSION("environment.conda.min.version", "4.0.1"),
```

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/environment/EnvironmentId.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package org.apache.submarine.server.api.environment;
+
+import org.apache.submarine.commons.utils.AbstractUniqueIdGenerator;
+
+/**
+ * The unique id for environment. Formatter:
+ * environment_${server_timestamp}_${counter} Such as:
+ * environment__1577627710_0001
```
jiwq, line 27:
```suggestion
 * environment_1577627710_0001
```

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/EnvironmentSpec.java
```diff
@@ -0,0 +1,109 @@
...
+ * under the License.
+ */
+
+package org.apache.submarine.server.api.spec;
+
+/**
+ * Submarine environment spec.
+ */
+/**
+ * @author mani
```
jiwq, line 26: Should remove the author info.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/KernelSpec.java
```diff
@@ -0,0 +1,49 @@
+package org.apache.submarine.server.api.spec;
```
jiwq, line 1: Missing the header license.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/KernelSpec.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.List;
+
+/**
+ *
```
jiwq, line 6: Redundant blank line

submarine-server/server-core/pom.xml
```diff
@@ -423,7 +423,22 @@
       </exclusions>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+    	<groupId>org.apache.httpcomponents</groupId>
+    	<artifactId>httpclient</artifactId>
+    	<version>${httpclient.version}</version>
+    	<exclusions>
+        	<exclusion>
```
jiwq, line 431: The indent should keep two space.

manirajv06:
> @manirajv06 Thanks for your contribution. Sorry for the later reply, some nits should fix:
> 
> * Missing the header license
> * Indentation format needs to be consistent
> * Should remove the redundant blank lines
> * I not see the update about `swagger` in `Bootstrap.java` which provide the `open api` service.

Fixed all review comments.

jiwq:
+LGTM Will merge if no more comments. Thanks @manirajv06 for the contribution.

wangdatan:
+1 from my side too, @manirajv06 , can you create a follow-up ticket to add documentation, I will follow the doc and try it and give more feedbacks. Thanks for your contribution!

manirajv06:
CI failures are not related.

manirajv06:
Taken care

manirajv06:
> Hi @manirajv06 could you help to check the CI failure? And the `0.5.0` version should replace with `0.5.0-SNAPSHOT` in `Bootstrap.java` file.

Taken care

liuxunorg:
@manirajv06 Please fixed travis ci fail.
https://travis-ci.org/github/apache/submarine/builds/705069327

image


```
Failed tests: 
  DictAnnotationTest>CommonDataTest.startTest:66->CommonDataTest.createAllTable:89->CommonDataTest.createDept:110->CommonDataTest.assertDeptResponseSuccess:245 null
Tests in error: 
  TeamServiceTest.queryPageList:63 Â»  org.apache.ibatis.exceptions.PersistenceEx...
```

manirajv06:
@liuxunorg I checked those failures and don't think failures are related to this change.

submarine-server/server-core/src/main/java/org/apache/submarine/server/database/utils/MyBatisUtil.java
```diff
@@ -34,33 +34,39 @@
...
+  private static SqlSessionFactory metastoreSqlSessionFactory;
 
   static {
+    SubmarineConfiguration conf = SubmarineConfiguration.getInstance();
+    sqlSessionFactory = buildSqlSessionFactory("mybatis-config.xml",
+        conf.getJdbcUrl(), conf.getJdbcUserName(), conf.getJdbcPassword());
+    metastoreSqlSessionFactory = buildSqlSessionFactory(
+        "mybatis-config-metastore.xml", conf.getMetastoreJdbcUrl(),
+        conf.getMetastoreJdbcUserName(), conf.getMetastoreJdbcPassword());
+  }
```
liuxunorg, line 46: @manirajv06 I guess that travis failure may be related to this.
Or where you have modified other database operations.

liuxunorg:
@manirajv06 If travis fails, this PR cannot be merged. Please check the code.
In addition, I give you a suggestion, the scope of the modification of each PR can be smaller in the future, so that it is easier to merge, and your PR content is too much. So it will take a lot of time to review.

submarine-server/server-core/src/main/java/org/apache/submarine/server/database/utils/MyBatisUtil.java
manirajv06: I spent some more time on these issues again to get more details. There are 2 issues:

1. https://travis-ci.org/github/apache/submarine/jobs/705069335

==>  Preparing: INSERT INTO sys_dict(`id`, `dict_code`, `dict_name`, `description`, `create_by`, `create_time`, `update_time`) VALUES ( REPLACE(UUID(), '-', ''), ?, ?, ?, ?, now(), now()) 
==> Parameters: SYS_USER_SEX(String), name(String), description(String), null
[ERROR] 2020-07-06 02:03:29,612 method:org.apache.submarine.server.workbench.rest.SysDictRestApi.add(SysDictRestApi.java:107)
### Error updating database.  Cause: com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Duplicate entry 'SYS_USER_SEX' for key 'UK_SYS_DICT_DICT_CODE'
### The error may involve org.apache.submarine.server.workbench.database.mappers.SysDictMapper.insertSysDict-Inline
### The error occurred while setting parameters
### SQL: INSERT INTO sys_dict(`id`, `dict_code`, `dict_name`, `description`,                 `create_by`, `create_time`, `update_time`)     VALUES (               REPLACE(UUID(), '-', ''),             ?, ?, ?,                 ?, now(), now())
### Cause: com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Duplicate entry 'SYS_USER_SEX' for key 'UK_SYS_DICT_DICT_CODE'

I can able to reproduce the error in my master branch itself. SysUserRestApiTest extends CommonDataTest class and CommonDataTest#startTest (@BeforeClass) tries to load the data. docs/submarine-data.sql also populates the same data. Seems to be duplicate entry issue. (or) Am I missing anything?

2. https://travis-ci.org/github/apache/submarine/jobs/705069329

### Error querying database.  Cause: java.sql.SQLException: Error setting driver on UnpooledDataSource. Cause: java.lang.ClassNotFoundException: Cannot find class: com.mysql.jdbc.Driver
### The error may exist in org/apache/submarine/database/mappers/SysUserMapper.xml
### The error may involve org.apache.submarine.server.workbench.database.mappers.SysUserMapper.login
### The error occurred while executing a query
### Cause: java.sql.SQLException: Error setting driver on UnpooledDataSource. Cause: java.lang.ClassNotFoundException: Cannot find class: com.mysql.jdbc.Driver
org.apache.ibatis.exceptions.PersistenceException: 
### Error querying database.  Cause: java.sql.SQLException: Error setting driver on UnpooledDataSource. Cause: java.lang.ClassNotFoundException: Cannot find class: com.mysql.jdbc.Driver
### The error may exist in org/apache/submarine/database/mappers/SysUserMapper.xml
### The error may involve org.apache.submarine.server.workbench.database.mappers.SysUserMapper.login
### The error occurred while executing a query
### Cause: java.sql.SQLException: Error setting driver on UnpooledDataSource. Cause: java.lang.ClassNotFoundException: Cannot find class: com.mysql.jdbc.Driver

MySQL driver class is missing...Seems to be intermittent issue?

liuxunorg: @kevin85421 Can you help to see this problem?

manirajv06: @kevin85421 Had a chance to look at this?

pingsutw: @manirajv06 Thanks for the contribution.
I'm going to help you fix the test failure.
After I find the root causes, I will merge the PR firstly, then I will fix the test failure in the next PR. :)

pingsutw:
Thanks all.
I've fixed the test failure in #347

wangdatan:
@manirajv06 thank you so much for contributing this! @pingsutw / @tangzhankun / @jiwq / @liuxunorg thanks for reviewing the patch.

PR #309, opened 2020-06-10
SUBMARINE-524. Add travis test of commons-unixusersync

Eroschang:
### What is this PR for?
To test the module of commons-unixusersync.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-524

https://travis-ci.org/github/Eroschang/submarine

image

### Questions:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -170,6 +170,18 @@ matrix:
         - TEST_MODULES="-pl org.apache.submarine:submarine-commons-runtime"
         - TEST_PROJECTS=""
 
+    - name: Test submarine commons-unixusersync
+      language: java
+      jdk: openjdk8
+      dist: xenial
+      env:
+        - PROFILE="-Phadoop-2.9"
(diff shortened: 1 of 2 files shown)


tangzhankun:
@Eroschang Thanks for the contribution. One doubt, why the commons-unixusersync only have test cases now? Do we need nny other utilities of LDAP APIs?

liuxunorg:
> @Eroschang Thanks for the contribution. One doubt, why the commons-unixusersync only have test cases now? Do we need nny other utilities of LDAP APIs?

Because synchronous ldap can be achieved through the ranger's exist modules,
But now I and @Eroschang  don't know where to put the functions of `commons-unixusersync`, so I have a separate module.

We have implemented the function first in this way, and we can easily adjust it when the requirements are clarified later.

PR #317, opened 2020-06-18
SUBMARINE-529. Change apiVersion

wang0630:
### What is this PR for?
k8s installed by kind 0.8.0 seems that it does not support apiVersion apps/v1beta1 anymore.

Change to apps/v1 to build successfully.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-529

https://travis-ci.com/github/wang0630/submarine/builds/172083779

### Questions:

### DIFF
--- a/helm-charts/submarine/templates/submarine-database.yaml
+++ b/helm-charts/submarine/templates/submarine-database.yaml
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-apiVersion: apps/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: "{{ .Values.submarine.database.name }}"


tangzhankun:
@wang0630 thanks for the update. The`Deployment`'s `apiVersion: apps/v1` is depending on the k8s version instead of kind. "Kind" is just a testing tool.
IIRC, from v1.16, the "apps/v1beta1" is deprecated. It's better that we improve our helm charts to support installation on all k8s version. If no possible, I'm fine with this PR in first.

wang0630:
> The `deployment` resource supports `apps/v1` version in v1.14 and the minimum version number of K8s submarine supports is v1.14. So this PR LGTM, if no more comments I will merge it later.

Noted with thanks!

PR #323, opened 2020-06-23
SUBMARINE-542. [SDK] get_log error when experiment is not started

pingsutw:
### What is this PR for?
Fix get_log error when the experiment is not started
```bash
IndexError                                Traceback (most recent call last)
<ipython-input-13-8bf55eff97d1> in <module>
----> 1 submarine_client.get_log(id)~/opt/anaconda3/envs/python3-7/lib/python3.7/site-packages/submarine/experiment/api/experiment_client.py in get_log(self, id, master)
    124 
    125         if master is True:
--> 126             log_contents = [log_contents[0]]
    127 
    128         for log_content in log_contents:IndexError: list index out of range
```

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-542

https://travis-ci.org/github/pingsutw/hadoop-submarine/builds/701089803

### Questions:

### DIFF
--- a/submarine-sdk/pysubmarine/submarine/experiment/api/experiment_client.py
+++ b/submarine-sdk/pysubmarine/submarine/experiment/api/experiment_client.py
@@ -125,7 +125,7 @@ class ExperimentClient:
         response = self.experiment_api.get_log(id)
         log_contents = response.result['logContent']
 
-        if master is True and len(log_contents) !=0:
+        if master is True and len(log_contents) != 0:
             log_contents = [log_contents[0]]
 
         for log_content in log_contents:
--- a/submarine-sdk/pysubmarine/submarine/experiment/api/experiment_client.py
+++ b/submarine-sdk/pysubmarine/submarine/experiment/api/experiment_client.py
@@ -114,17 +114,18 @@ class ExperimentClient:
         return response.result
 
     def get_log(self, id, onlyMaster=False):
-        """  
+        """
         Get training logs of all pod of the experiment.
         By default get all the logs of Pod
         :param id: experiment id
-        :param onlyMaster: By default include pod log of "master" which might be Tensorflow PS/Chief or PyTroch master
+        :param onlyMaster: By default include pod log of "master" which might be
+         Tensorflow PS/Chief or PyTorch master
         :return: str: pods logs
         """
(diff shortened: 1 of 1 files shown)


submarine-sdk/pysubmarine/submarine/experiment/api/experiment_client.py
```diff
@@ -110,19 +113,19 @@ def delete_experiment(self, id):
         response = self.experiment_api.delete_experiment(id)
         return response.result
 
-    def get_log(self, id, master=True):
+    def get_log(self, id, master=False):
         """
         Get training logs of the experiment.
         By default only get the logs of Pod that has labels 'job-role: master'.
-        :param master: By default get pod with label 'job-role: master' pod if True.
+        :param master: Get pod with label 'job-role: master' pod if True.
```
tangzhankun, line 120: The description of "master" param seems confusing. I thought get_log(id) will get all pod's log, and get_log(id, master=false) will discard the master pod log.
But per the description, get_log(id, master=True) will get a master log. This is confusing.

> Get pod with label 'job-role: master' pod if True.

pingsutw: I just follow the convention in `tf-operator` 
https://github.com/kubeflow/tf-operator/blob/cf6ebd6d2d9f679646dad3530e9314f63505f98a/sdk/python/kubeflow/tfjob/api/tf_job_client.py#L356-L371
Maybe we could change the description, so it will be more clear for the users.
Any suggestion for the comments and param in `get_log`

tangzhankun: How about this?
```Python
"""
get_logs(self, id, onlyMaster=False)   
Get training logs of all pod of the experiment.
By default get all the logs of Pod
:param id: experiment id
:param onlyMaster: By default include pod log of "master" which might be Tensorflow PS/Chief or PyTroch master
:return: str: pods logs
"""
```

pingsutw: Thanks, @tangzhankun.
it's more clear than before. I've updated it.

PR #348, opened 2020-07-15
SUBMARINE-565. Documentation for Environment Management REST API

manirajv06:
### What is this PR for?
Documentation for Environment Management REST API's

### What type of PR is it?
Documentation

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-565

### How should this be tested?
* First time? Setup Travis CI as described on https://submarine.apache.org/contribution/contributions.html#continuous-integration
* Strongly recommended: add automated unit tests for any new or changed behavior
* Outline any manual steps to test the PR here.

### Questions:
* Does the licenses files need update? Yes/No
* Is there breaking changes for older versions? Yes/No
* Does this needs documentation? Yes/No

PR #362, opened 2020-07-26
SUBMARINE-570. Support run experiment/notebook with synced code

manirajv06:
### What is this PR for?
https://issues.apache.org/jira/browse/SUBMARINE-570

### What type of PR is it?
Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-570

### Questions:

### DIFF
--- a/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
+++ b/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
@@ -30,6 +30,7 @@ public abstract class AbstractCodeLocalizer implements CodeLocalizer {
   public static final String CODE_LOCALIZER_PATH = "/code";
   public static final String CODE_LOCALIZER_MOUNT_NAME = "code-dir";
   public static final String CODE_LOCALIZER_INIT_CONTAINER_NAME = "code-localizer";
+  public static final String CODE_LOCALIZER_PATH_ENV_VAR = "CODE_PATH";
   private String url;
   
   public AbstractCodeLocalizer(String url) {
--- a/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
+++ b/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
@@ -1,0 +1,54 @@
+package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
+
+import org.apache.submarine.server.api.spec.ExperimentSpec;
+import org.apache.submarine.server.api.spec.ExperimentTaskSpec;
+
+import io.kubernetes.client.models.V1Container;
+import io.kubernetes.client.models.V1EmptyDirVolumeSource;
+import io.kubernetes.client.models.V1PodSpec;
+import io.kubernetes.client.models.V1Volume;
+import io.kubernetes.client.models.V1VolumeMount;
+
+public abstract class AbstractCodeLocalizer implements CodeLocalizer {
+
+  protected ExperimentSpec experimentSpec;
+  
+  public AbstractCodeLocalizer(ExperimentSpec experimentSpec) {
+    this.experimentSpec = experimentSpec;
+  }
+  
+  @Override
+  public void localize(V1PodSpec podSpec) {
+    
+    V1Volume volume = new V1Volume();
+    volume.setName("code-dir");
+    volume.setEmptyDir(new V1EmptyDirVolumeSource());
+    podSpec.addVolumesItem(volume);
+    
+  }
+
+  public static CodeLocalizer getCodeLocalizer(ExperimentSpec experimentSpec) {
+
+    String syncMode = experimentSpec.getCode().getSyncMode();
+    if (syncMode.equals(CodeLocalizerModes.GIT.getMode())) {
+      return GitCodeLocalizer.getGitCodeLocalizer(experimentSpec);
+    } else {
+      return new DummyCodeLocalizer(experimentSpec);
+    }
+  }
+
+  public enum CodeLocalizerModes {
+
+    GIT("git"), HDFS("hdfs"), NFS("nfs"), S3("s3");
+
+    private final String mode;
+
+    CodeLocalizerModes(String mode) {
+      this.mode = mode;
+    }
+
+    public String getMode() {
+      return this.mode;
+    }
+  }
+}
--- a/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
+++ b/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
@@ -19,7 +19,7 @@
 
 package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
 
-import org.apache.submarine.server.api.spec.ExperimentSpec;
+import org.apache.submarine.server.api.exception.InvalidSpecException;
 
 import io.kubernetes.client.models.V1EmptyDirVolumeSource;
 import io.kubernetes.client.models.V1PodSpec;
@@ -27,29 +27,33 @@ import io.kubernetes.client.models.V1Volume;
 
 public abstract class AbstractCodeLocalizer implements CodeLocalizer {
 
-  protected ExperimentSpec experimentSpec;
+  private String url;
   
-  public AbstractCodeLocalizer(ExperimentSpec experimentSpec) {
-    this.experimentSpec = experimentSpec;
+  public AbstractCodeLocalizer(String url) {
+    this.url = url;
+  }
+  
+  /**
+   * @return the url
+   */
+  public String getUrl() {
+    return url;
   }
   
   @Override
   public void localize(V1PodSpec podSpec) {
-    
     V1Volume volume = new V1Volume();
     volume.setName("code-dir");
     volume.setEmptyDir(new V1EmptyDirVolumeSource());
     podSpec.addVolumesItem(volume);
-    
   }
 
-  public static CodeLocalizer getCodeLocalizer(ExperimentSpec experimentSpec) {
-
-    String syncMode = experimentSpec.getCode().getSyncMode();
+  public static CodeLocalizer getCodeLocalizer(String syncMode, String url)
+      throws InvalidSpecException {
     if (syncMode.equals(CodeLocalizerModes.GIT.getMode())) {
-      return GitCodeLocalizer.getGitCodeLocalizer(experimentSpec);
+      return GitCodeLocalizer.getGitCodeLocalizer(url);
     } else {
-      return new DummyCodeLocalizer(experimentSpec);
+      return new DummyCodeLocalizer(url);
     }
   }
 
--- a/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
+++ b/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
@@ -1,3 +1,22 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
 package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
 
 import org.apache.submarine.server.api.spec.ExperimentSpec;
--- a/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
+++ b/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
@@ -1,13 +1,10 @@
 package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
 
 import org.apache.submarine.server.api.spec.ExperimentSpec;
-import org.apache.submarine.server.api.spec.ExperimentTaskSpec;
 
-import io.kubernetes.client.models.V1Container;
 import io.kubernetes.client.models.V1EmptyDirVolumeSource;
 import io.kubernetes.client.models.V1PodSpec;
 import io.kubernetes.client.models.V1Volume;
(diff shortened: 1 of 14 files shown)


manirajv06:
@wangdatan First cut implementation. Can you please take a look?
 Need to do integration testing to ensure code has been synced from git repos etc.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/GitCodeLocalizer.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public void localize(V1PodSpec podSpec) {
+    super.localize(podSpec);
+  }
+
+  public GitCodeLocalizer(ExperimentSpec experimentSpec) {
+    super(experimentSpec);
+  }
+
+  public static CodeLocalizer getGitCodeLocalizer(
```
wangdatan, line 36: Instead of get CodeInitalizer from ExperimentSpec, it's better to use URL to get the CodeLocalizer.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/HTTPGitCodeLocalizer.java
```diff
@@ -0,0 +1,59 @@
...
+  
+  @Override
+  public void localize(V1PodSpec podSpec) {
+    
+    V1Container container = new V1Container();
+    
+    container.setName("git-localizer");
+    container.setImage("git-sync");
+    
+    V1VolumeMount mount = new V1VolumeMount();
```
wangdatan, line 45: Can we (or is it possible) make the volume mount logic common to different code initializers? 

And SSHGitCode.. can share this implementation, correct?

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/GitCodeLocalizer.java
```diff
@@ -0,0 +1,64 @@
...
+ * under the License.
+ */
+
+package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
+
+import org.apache.submarine.server.api.spec.ExperimentSpec;
+
+import io.kubernetes.client.models.V1PodSpec;
+
+public abstract class GitCodeLocalizer extends AbstractCodeLocalizer {
```
wangdatan, line 26: How we handle Git pull credentials, should it part of ExperimentSpec, or part of Namespace secret. 

@manirajv06, @liuxunorg , @tangzhankun: any ideas here?

docs/userdocs/k8s/api/experiment.md
```diff
@@ -167,6 +167,79 @@ Above example assume environment "my-submarine-env" already exists in Submarine.
 }
 ```
 
+## Create Experiment (Using Pre-defined/Stored Environment)
```
wangdatan, line 170: I'm not sure why we make this change, I thought it will include /code init spec, but it didn't. Do you plan to add it later?

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
```diff
@@ -0,0 +1,70 @@
...
+
+package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
+
+import org.apache.submarine.server.api.spec.ExperimentSpec;
+
+import io.kubernetes.client.models.V1EmptyDirVolumeSource;
+import io.kubernetes.client.models.V1PodSpec;
+import io.kubernetes.client.models.V1Volume;
+
+public abstract class AbstractCodeLocalizer implements CodeLocalizer {
```
wangdatan, line 28: I think we should remove all experimentSpec from CodeLocalizer (and its implementations). 
One important purpose is to allow use CodeSpec in both ExperimentSpec and Notebook. Even Notebook is not ready yet, I think the implementation can general enough to support both. 

Regarding the implementation, we should have no problem to remove experimentSpec from implementation, what we only need to do is use code URL to init codeLocalizer.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/GitCodeLocalizer.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public GitCodeLocalizer(ExperimentSpec experimentSpec) {
+    super(experimentSpec);
+  }
+
+  public static CodeLocalizer getGitCodeLocalizer(
+      ExperimentSpec experimentSpec) {
+
+    String url = experimentSpec.getCode().getUrl();
```
wangdatan, line 39: We should parse URL properly and check the protocol instead of using .contains: 
https://docs.oracle.com/javase/tutorial/networking/urls/urlInfo.html

wangdatan: And we should check URL for code spec is valid, if it is not valid, we should fail submission of the experiment/notebook.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/HTTPGitCodeLocalizer.java
manirajv06: Ah ! Have followed the same approach for VolumeDevice but missed this. Taken care.

docs/userdocs/k8s/api/experiment.md
manirajv06: This change shouldn't be part of this PR. It is part of https://github.com/apache/submarine/pull/348. Will check again.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/GitCodeLocalizer.java
manirajv06: Probably, one time activity based approach is preferable as it doesn't change for every experiment. It remains same for repo's.

manirajv06:
@wangdatan @liuxunorg Forgot to mention that code has been tested using e2e tests. Was able to see repo code has been synced to /code directory inside k8 container.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/HTTPGitCodeLocalizer.java
```diff
@@ -0,0 +1,35 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.submarine.server.submitter.k8s.experiment.codelocalizer;
+
+import io.kubernetes.client.models.V1PodSpec;
+
+public class HTTPGitCodeLocalizer extends GitCodeLocalizer {
```
wangdatan, line 24: @manirajv06 , it looks like HTTP and SSH Git localizer pretty same, I'm wondering why we need the two implementations. And why `localize` of `SSHGitCodeLocalizer` is empty?

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
```diff
@@ -168,8 +171,35 @@ private static V1PodTemplateSpec parseTemplateSpec(
...
+     */
+    if (experimentSpec.getCode() != null) {
+      CodeLocalizer localizer = AbstractCodeLocalizer.getCodeLocalizer(
+          experimentSpec.getCode().getSyncMode(),
+          experimentSpec.getCode().getUrl());
+      localizer.localize(podSpec);
+
+      if (podSpec.getInitContainers() != null
+          && podSpec.getInitContainers().size() > 0) {
+        String volumeName = podSpec.getInitContainers().get(0).getVolumeMounts()
```
wangdatan, line 186: Is this part of logic generic enough to support multiple volumes? (It is possible in the future we want to support mount other volumes to the pod, such as an NFS volume. Instead of get(0), can we get the volume by dir name?

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
```diff
@@ -0,0 +1,74 @@
...
+   * @return the url
+   */
+  public String getUrl() {
+    return url;
+  }
+  
+  @Override
+  public void localize(V1PodSpec podSpec) {
+    V1Volume volume = new V1Volume();
+    volume.setName("code-dir");
```
wangdatan, line 46: can we make `code-dir` to be a constant and can be referred in other places (like test)?

docs/userdocs/k8s/api/experiment.md
manirajv06: Taken care

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/HTTPGitCodeLocalizer.java
manirajv06: As of today, it seem like there are no differences. But when we get into "git ssh mode" implementation, we will need this separation. For ex, GIT_SYNC_SSH needs to be set as "true".

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
manirajv06: (Yes, pod containers may mount multiple volumes in future for various reasons which can be pat of other init container process). But at any point of time, this "code localizer" init container would have only 1 volume mount. However, open to change if you think otherwise.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
manirajv06: Taken care.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/SSHGitCodeLocalizer.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class SSHGitCodeLocalizer extends GitCodeLocalizer {
+
+  public SSHGitCodeLocalizer(String url) {
+    super(url);
+  }
+
+  @Override
+  public void localize(V1PodSpec podSpec) {
+  }
```
jiwq, line 32: @manirajv06 Thanks for your great contribution. Can you mind add TODO comment in here?

manirajv06: @jiwq Scope of this PR/JIRA is to provide support http based git sync mode. Added this class now to get clear picture of how interfaces and its impl look like. More over, we will be adding support for ssh very soon. Will create a jira for the same. cc @wangdatan

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/HTTPGitCodeLocalizer.java
wangdatan: gotcha, make sense.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
wangdatan: It makes sense to me. We can revisit this once we want to support multiple volumes.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/SSHGitCodeLocalizer.java
manirajv06: @jiwq Added.

manirajv06:
@wangdatan 

Since PR has taken a good shape and to bring conclusion, writing down plans to close gaps:

1. How do we handle private git repo to sync the code?

First cut:

Can we do the following in experiment spec?

spec {
..
code: {
"mode": "git",
"url": "..",
"username": "..",
"password": ".."
}
..
}

and use the credentials to sync the code from git. It carries the risk of showing above info in pod description (kubectl describe pod ..), in addition to expecting username and password in plain text as part of experiment spec.

Next Phase:

Creating a secret in K8 and using the same for code sync is much cleaner and better solution. However, we need username/password info at least one time to create a "secret" in K8 internally. Once this is done, we can use the same secret when same repo has been used in any experiment and pass those credentials to "code localizer" init container to sync the code. This way, there is no need of passing the username/password again and again. 

We can bring this change in intuitive way for Note Book users as it is Web UI driven. Where as, users coming through API's would find it bit confusing. For example, Users pass username/password as part of experiment spec in first run and don't bother to remove those creds from second run onwards. I think we can debate more on this before concluding the approach.

Thoughts?

2. How localized code would be used?

My understanding is, using commands mentioned in "cmd" of "meta" spec.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/GitCodeLocalizer.java
tangzhankun: The credential seems should be created in the workbench and couple with the user account. And we may also think about the refresh of the secure token.  I'd prefer to not do the secure localizer for now.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/experiment/codelocalizer/AbstractCodeLocalizer.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.apache.submarine.server.api.exception.InvalidSpecException;
+
+import io.kubernetes.client.models.V1EmptyDirVolumeSource;
+import io.kubernetes.client.models.V1PodSpec;
+import io.kubernetes.client.models.V1Volume;
+
+public abstract class AbstractCodeLocalizer implements CodeLocalizer {
+
+  public static final String CODE_LOCALIZER_PATH = "/code";
```
tangzhankun, line 30: Do we need to set this environment variable to the other containers so that the user can set command like
"python $CODE_PATH/src/train.py"?

liuxunorg:
This PR has been too long, we can merge it in first, and then improve if there are problems.
What do you think?

manirajv06:
@tangzhankun I am also having similar thoughts and to take care of private repo's in separate jira. Will create a jira.

@liuxunorg Made final commit for setting env var to bring logical conclusion. It is in good shape.

tangzhankun:
LGTM. Thanks @manirajv06 . Will merge it if no more comments.

PR #378, opened 2020-08-13
SUBMARINE-599. [WEB] Clone experiment through UI

pingsutw:
# What is this PR for?
Add a button for the user to clone an experiment. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-599

https://travis-ci.org/github/pingsutw/hadoop-submarine/builds/717569342

Screenshot from 2020-08-13 20-21-51

Submarine - experiment

### Questions:

### DIFF
--- a/submarine-workbench/workbench-web-ng/package.json
+++ b/submarine-workbench/workbench-web-ng/package.json
@@ -27,6 +27,7 @@
     "lint-staged": "^10.2.2",
     "lodash": "^4.17.19",
     "md5": "^2.2.1",
+    "nanoid": "^3.1.12",
     "ng-zorro-antd": "8.1.2",
     "ngx-monaco-editor": "^8.0.0",
     "rxjs": "~6.4.0",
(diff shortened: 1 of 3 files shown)


pingsutw:
@jiwq I've updated a screenshot.
The purpose of cloning the experiment is to run a new experiment that is similar to the previous one.
Sometimes we will run many different experiments with different CPU, memory, or model parameter.
Instead of retyping the whole spec (meta, env, replica, command, ..etc), we could just clone an experiment and then modify CPU or memory number, and launch a new experiment.

jiwq:
> @jiwq I've updated a screenshot.
> The purpose of cloning the experiment is to run a new experiment that is similar to the previous one.
> Sometimes we will run many different experiments with different CPU, memory, or model parameter.
> Instead of retyping the whole spec (meta, env, replica, command, ..etc), we could just clone an experiment and then modify CPU or memory number, and launch a new experiment.

I see, thanks for your introduce and contribution.  +1 If no more comments, I'll commit this.

pingsutw:
@jiwq Thanks for the review.

PR #426, opened 2020-10-03
SUBMARINE-652. Autoformat staged files before committing

ByronHsu:
### What is this PR for?
Although we applied prettier in the workbench. However, even if the code format violates the rule, we can still commit the code.
In that way, prettier might become useless, so we should autoformat the file in precommit.

Therefore, I use husky + lint-staged to construct a mechanism which can automatically format only staged file in git, which is quite efficient and convenient.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-652

https://travis-ci.org/github/ByronHsu/submarine/builds/732470123

Screen Shot 2020-10-03 at 4 01 01 PM

### Questions:

### DIFF
--- a/submarine-workbench/workbench-web-ng/package.json
+++ b/submarine-workbench/workbench-web-ng/package.json
@@ -9,7 +9,7 @@
     "test": "ng test",
     "lint": "ng lint",
     "e2e": "ng e2e",
-    "autoformat": "prettier --write --trailing-comma none --single-quote 'src/**/*.{js,json,ts,html}'",
+    "autoformat": "prettier --config .prettierrc --write 'src/**/*.{js,json,ts,html}'",
     "webdriver": "webdriver-manager update --ignore_ssl = true"
   },
   "private": true,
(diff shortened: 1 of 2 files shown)


submarine-workbench/workbench-web-ng/package.json
```diff
@@ -49,14 +53,16 @@
...
     "jasmine-core": "~3.4.0",
     "jasmine-spec-reporter": "~4.2.1",
     "karma": "~4.1.0",
     "karma-chrome-launcher": "~2.2.0",
-    "karma-firefox-launcher": "~1.3.0",
     "karma-coverage-istanbul-reporter": "~2.0.1",
+    "karma-firefox-launcher": "~1.3.0",
     "karma-jasmine": "~2.0.1",
     "karma-jasmine-html-reporter": "^1.4.0",
+    "lint-staged": "^10.4.0",
```
kevin85421, line 65: Thank you for your contribution! I am wondering why we need to remove "lint-staged" from "dependencies" and add it to "devDependencies"?

ByronHsu: Because code formatter is only used in the development stage. In production, we do not need to run formatter.

ByronHsu: lint-staged is a package by which you can some check (in here, is prettier) to staged files.

kevin85421:
LGTM

ByronHsu:
@pingsutw Can you please review this PR? Thanks!

pingsutw (approved):
LGTM, thanks

xunliu (approved):

PR #433, opened 2020-10-09
SUBMARINE-656. Fix the wrong message when creating k8s object fail

aeioulisa:
### What is this PR for?
It always shows "Internal Server Error" when creating k8s object fails.
Now it can show the correct k8s error message.
 
### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-656

https://travis-ci.org/github/aeioulisa/submarine/builds/734180722

### Questions:
* Does this need documentation? No


submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/K8sSubmitter.java
```diff
@@ -126,7 +127,7 @@ public Experiment createExperiment(ExperimentSpec spec) throws SubmarineRuntimeE
       throw new SubmarineRuntimeException(400, e.getMessage());
     } catch (ApiException e) {
       LOG.error("K8s submitter: parse Job object failed by " + e.getMessage(), e);
-      throw new SubmarineRuntimeException(e.getCode(), e.getMessage());
+      throw new SubmarineRuntimeException(400, "K8s submitter: parse Job object failed by " + e.getMessage());
```
jiwq, line 130: Why change the code to 400? I think should use the code responded by upstream.

aeioulisa: The code will be transformed into "javax.ws.rs.core.Response.Status" at "JsonResponse.java", but there's no corresponding code here. This will lead to another error(java.lang.IllegalArgumentException: Response status must not be 'null'). This new error will be returned as response so that the user does not get the correct error message.

jiwq: Okay, we can fix it in http response rather than here. Any thoughts?

xunliu (approved):
LGTM

PR #437, opened 2020-10-20
SUBMARINE-663. Use md5sum command in publish_release.sh

pingsutw:
### What is this PR for?
md5 could only run on macOS machines.

Add md5sum, so that we could also run on Linux machines

ref: https://github.com/apache/spark/blame/688d016c7acc4b9d96d75b40123be9f40b7b2693/dev/create-release/release-build.sh#L444-L452

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-663

https://travis-ci.org/github/pingsutw/hadoop-submarine/builds/737266426

### Questions:


xunliu (approved):
LGTM

PR #438, opened 2020-10-20
SUBMARINE-611. Fix a bug in add new project page

brayce1996:
### What is this PR for?
"Create new project page" UI makes the "Next Step button" become clickable before all required fields are filled.
Now the button would stay unclickable until users filled all required fields.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-611

https://travis-ci.org/github/brayce1996/submarine/builds/736928222

1. Visibility is not filled, so Next Step button is disabled.
image
2. Visibility is filled, but public visibility requires Permission to be filled. Next Step button is disabled.
image
3. All required fields are filled, so Next Step button is enabled.
image

### Questions:


xunliu (commented):
LGTM
@brayce1996 Thank you fixed this problem.

PR #440, opened 2020-10-26
SUBMARINE-657. The problem of creating a new user

charlie0220:
### What is this PR for?
1. Correct the bug while creating an user without entering birthday.
Now if users don't enter the birthday, the default birthday will be null.

2. Add md5 encryption for the password while creating new user.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
[SUBMARINE-657](https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-657?filter=allopenissues)

[Travis-CI](https://travis-ci.com/github/charlie0220/submarine)

### Questions:

### DIFF
--- a/submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
+++ b/submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
@@ -151,12 +151,7 @@ export class UserDrawerComponent implements OnInit, OnChanges {
           delete formData['sex@dict'];
           delete formData['status@dict'];
 
-          if (formData.birthday) {
-            formData.birthday = format(formData.birthday, 'yyyy-MM-dd HH:mm:ss');
-          }
-          else{
-            formData.birthday = format(new Date(2000,0O0,0O1),'yyyy-MM-dd HH:mm:ss');
-          }
+          formData.birthday = formData.birthday ? format(formData.birthday, 'yyyy-MM-dd HH:mm:ss'): null;
 
           this.submit.emit(formData);
         }
--- a/submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
+++ b/submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
@@ -154,6 +154,9 @@ export class UserDrawerComponent implements OnInit, OnChanges {
           if (formData.birthday) {
             formData.birthday = format(formData.birthday, 'yyyy-MM-dd HH:mm:ss');
           }
+          else{
+            formData.birthday = format(new Date(2000,0O0,0O1),'yyyy-MM-dd HH:mm:ss');
+          }
 
           this.submit.emit(formData);
         }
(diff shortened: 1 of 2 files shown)


submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
```diff
@@ -153,6 +153,8 @@ export class UserDrawerComponent implements OnInit, OnChanges {
 
           if (formData.birthday) {
             formData.birthday = format(formData.birthday, 'yyyy-MM-dd HH:mm:ss');
+          }else{
```
pingsutw, line 156:
```suggestion
          } else {
```

pingsutw (commented):
Thanks @charlie0220 for the contribution.

submarine-workbench/workbench-web/src/app/services/user.service.ts
```diff
@@ -96,6 +96,7 @@ export class UserService {
 
   createUser(sysUser: Partial<SysUser>): Observable<SysUser> {
     const apiUrl = this.baseApi.getRestApi('/sys/user/add');
+    sysUser['password'] = md5(sysUser.password);
```
xunliu, line 99: The user password stored in the database is already encrypted.

submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
```diff
@@ -153,6 +153,8 @@ export class UserDrawerComponent implements OnInit, OnChanges {
 
           if (formData.birthday) {
             formData.birthday = format(formData.birthday, 'yyyy-MM-dd HH:mm:ss');
+          } else {
+            formData.birthday = format(new Date(2000,0O0,0O1),'yyyy-MM-dd HH:mm:ss');
```
xunliu, line 157: What time is `Date(2000,0O0,0O1)`?

submarine-workbench/workbench-web/src/app/services/user.service.ts
charlie0220: It seems like the password is not encrypted while creating a new user.
password

submarine-workbench/workbench-web/src/app/pages/workbench/manager/user-drawer/user-drawer.component.ts
charlie0220: I have updated this file. Now if user do not enter the birthday, the value stored into database will be null.

xunliu: Okay, What we should do is to encrypt the password in the database.

lowc1012 (commented):
@charlie0220 Thank you for your contribution.
I test this patch, and it works for me.

submarine-workbench/workbench-web/src/app/services/user.service.ts
xunliu, line 99: @charlie0220 When creating a user, you need to encrypt the user's password.
When the user logs in, the user’s password needs to be encrypted, and then corrected with the encrypted password in the database.

xunliu (approved):
LGTM

PR #469, opened 2020-12-03
SUBMARINE-692. [web] Replace hand-crafted authguard with Angular builtin function

ByronHsu:
### What is this PR for?
1. Replace hand-crafted authguard with Angular builtin function
2. Remove notebook route module. Because there is only one path there, we don't need to use routing module

### What type of PR is it?
[mprovement]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-692

https://github.com/ByronHsu/submarine/actions/runs/398816623


pingsutw (approved):
LGTM, Thanks @ByronHsu

xunliu (approved):
LGTM

PR #486, opened 2021-01-04
SUBMARINE-704. Fix empty server logs in pod

ByronHsu:
### What is this PR for?
The logs of server pod are empty because they were not directed properly. I use to tee command to direct them to stdout to make the logs can be accessed by kubectl logs.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-704

Kapture 2021-01-04 at 21 16 38

### Questions:


helm-charts/submarine/templates/submarine-server.yaml
```diff
@@ -66,8 +66,6 @@ spec:
 
         image: "{{ .Values.submarine.server.image }}"
         imagePullPolicy: {{ .Values.submarine.server.imagePullPolicy }}
-        command: ["/bin/bash", "-c"]
-        args: ["/opt/submarine-current/bin/submarine.sh start getMysqlJar"]
```
ByronHsu, line 70: I deleted them because it has already been declared in server Dockerfile. They are redundant.

ByronHsu:
@pingsutw @xunliu please help me review this PR. Thanks!

bin/submarine.sh
```diff
@@ -55,4 +55,4 @@ if [[ ! -d "${SUBMARINE_LOG_DIR}" ]]; then
   $(mkdir -p "${SUBMARINE_LOG_DIR}")
 fi
 
-exec $JAVA_RUNNER $JAVA_OPTS -cp ${SUBMARINE_SERVER_CLASSPATH} ${SUBMARINE_SERVER_MAIN} "$@" >> "${SUBMARINE_SERVER_LOGFILE}" 2>&1
+exec $JAVA_RUNNER $JAVA_OPTS -cp ${SUBMARINE_SERVER_CLASSPATH} ${SUBMARINE_SERVER_MAIN} "$@" | tee -a "${SUBMARINE_SERVER_LOGFILE}" 2>&1
```
jiwq, line 58: Use `tee` copies the data to `STDOUT` ? Is it the really reason caused the the logs empty?

ByronHsu: 'kubectl logs' shows the stdout of the container in the pod. 
The cause of 'kubectl logs' to be empty is that the container only direct the output to the log file. However, if we want to show the output by 'kubectl logs', we should direct it to stdout too.

jiwq: LGTM. Thanks for your contribution. Will merge if no more comments.

PR #495, opened 2021-01-20
SUBMARINE-713. Fix workbench proxy target

ByronHsu:
### What is this PR for?
Since we have applied the ingress controller, the port of server is switching from 8080 to 32080, and we should revise it in proxy config file.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-713


pingsutw (approved):
LGTM, Thanks @ByronHsu

xunliu (approved):
@ByronHsu Thank you for contributions.

PR #505, opened 2021-02-04
SUBMARINE-726. Run qlib workflow in the notebook

pingsutw:
### What is this PR for?
Add a qlib example in jupyter notebook, so that users can directly run a qlib workflow when creating a notebook.

qlib example:
https://github.com/microsoft/qlib/blob/main/examples/workflow_by_code.ipynb
https://colab.research.google.com/github/microsoft/qlib/blob/main/examples/workflow_by_code.ipynb#scrollTo=t1kkNKGi3Lle

**Note**:
Currently, we don't support install pip package in the environment, so I pre-installed `pyqlib` in docker images. 
We only can install python packages that are in the conda channel. 

### What type of PR is it?
[Feature]

### Todos
* [ ] - Suppoprt pip install in Submarine Environment

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-726

https://www.travis-ci.com/github/pingsutw/hadoop-submarine/builds/215961270

image

### Questions:
* Does the license files need an update? No

### DIFF
--- a/dev-support/docker-images/jupyter/Dockerfile
+++ b/dev-support/docker-images/jupyter/Dockerfile
@@ -96,11 +96,18 @@ RUN git clone https://github.com/apache/submarine && \
     pip install submarine/submarine-sdk/pysubmarine && \
     conda install nodejs && \
     conda install -c conda-forge jupyterlab jupyterlab-git && \
-    jupyter lab build && \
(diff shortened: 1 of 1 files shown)


tangzhankun:
@pingsutw Thanks for the contribution. LGTM.

PR #506, opened 2021-02-04
SUBMARINE-732. Wrap tensorboard logdir into an environment variable

ByronHsu:
### What is this PR for?
It is inconvenient for users to manually specify the tensorboard logdir every time, so I wrap the real logdir path into an env variable (SUBMARINE_LOG_DIR). 
The usage is as the following:

```json
  "meta": {
    "name": "tensorflow-tensorboard-dist-mnist-13",
    "namespace": "default",
    "framework": "TensorFlow",
    "cmd": "python /var/tf_mnist/mnist_with_summaries.py --log_dir=$(SUBMARINE_LOG_DIR) --learning_rate=0.01 --batch_size=20",
    "envVars": {
      "ENV_1": "ENV1"
    }
  },
```

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Tensorboard user doc

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-732

### Questions:


pingsutw (commented):
@ByronHsu Thanks for the contribution. 
IMHO, It's better to use `SUBMARINE_` as the prefix for those env variables that Submarine provides clients to use.

ByronHsu:
@pingsutw Thanks. I have fixed it!

pingsutw (approved):
+1

PR #509, opened 2021-02-09
SUBMARINE-734. Open a new page when clicking tensorboard button

ByronHsu:
### What is this PR for?
Currently, the tensorboard button will redirect the user to the tensorboard web on the same page. However, it is more convenient if we can create an additional webpage for users. 

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-734

### Questions:


pingsutw (approved):
+1

PR #521, opened 2021-02-23
SUBMARINE-753. Submarine-security doesn't support 'show tables' filtering

HSJiChen:
### What is this PR for?
Fix the bug that 'show tables' doesn't work in submarine-security module.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-753 
Submarine-security doesn't support 'show tables' filtering

### Questions:


yaooqinn:
Thanks for your first contribution and welcome! Please fix the title according to the contributing guide, and add an unit test for the related change

yaooqinn:
@SJTUChenJi can you check that the ut here https://github.com/apache/submarine/blob/3041ef26ad04c0bddd2257a28694aa4e2b4cc837/submarine-security/spark-security/src/test/scala/org/apache/submarine/spark/security/AuthorizationTest.scala#L109 also passes before this PR? 

I am a lit confused

pingsutw:
@SJTUChenJi any update on this PR?

SJTUChenJi:
@pingsutw No, thanks!

PR #530, opened 2021-03-09
SUBMARINE-759. Update submarine-operator.md

kevin85421:
### What is this PR for?
The document of submarine-operator is outdated, and thus we cannot launch the submarine operator with this document. 
For example, 

(1) (submarine-operator.md) Kind v0.6.0 deprecates `kind get kubeconfig-path`
(2) (manifests/submarine-operator/deployment.yaml) `selector` field is necessary.
(3) (manifests/submarine-operator/deployment.yaml) Now, deployment is `apiVersion: apps/v1` rather than `extensions/v1beta1`.
(4) The .yaml files in `manifests/` are very different with the files in `helm-charts/`. Therefore, some errors will occur. For instance, there is no `configMap` in  `helm-charts/submarine/templates/submarine-server.yaml`, but a `configMap` is defined in `manifests/submarine-cluster/submarine-server.yaml`. 

Hence, we need to update this document, and we should refactor this component next.

### What type of PR is it?
[Documentation]

### Todos
* Refactor submarine-operator

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-759?filter=allopenissues

### Questions:


pingsutw (approved):
LGTM, thanks

PR #553, opened 2021-04-09
SUBMARINE-786. Use helm in k8s cluster

kevin85421:
### What is this PR for?
Because the helm chart of Submarine depends on many other charts, submarine-operator needs to be able to install these charts with helm.

### What type of PR is it?
[Feature]

### Todos
* Check if the release has already existed automatically

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-786?filter=allopenissues

* Out-of-cluster

https://user-images.githubusercontent.com/20109646/114165283-dc7bab80-995e-11eb-831d-e9e9d6574eba.mov

* In-cluster

https://user-images.githubusercontent.com/20109646/114165309-e2718c80-995e-11eb-95c3-c28c5d206d0b.mov

### Questions:


submarine-cloud-v2/controller.go
```diff
@@ -111,6 +112,34 @@ func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) error {
...
+		"svc",
+		"helm-install-example-release",
+		"default",
+		map[string]string {
+			"set": "ports[0].protocol=TCP,ports[0].port=80,ports[0].targetPort=9376",
+		},	
+	)
+
+	klog.Info("[Helm example] Sleep 60 seconds")
+	time.Sleep(time.Duration(60) * time.Second)
```
xunliu, line 136: Why does this need to sleep?

submarine-cloud-v2/pkg/helm/helm.go
```diff
@@ -0,0 +1,248 @@
...
+	"helm.sh/helm/v3/pkg/cli/values"
+	"helm.sh/helm/v3/pkg/downloader"
+	"helm.sh/helm/v3/pkg/getter"
+	"helm.sh/helm/v3/pkg/repo"
+	"helm.sh/helm/v3/pkg/strvals"
+)
+
+func HelmInstall(url string, repoName string, chartName string, releaseName string, namespace string, args map[string]string) (*action.Configuration){
+	var settings *cli.EnvSettings
+	os.Setenv("HELM_NAMESPACE", namespace)
```
xunliu, line 49: What is the scope of this enviroment variable `HELM_NAMESPACE`?
Will it affect other places?

xunliu (commented):
@kevin85421 Thanks you for your contribution. good job.

submarine-cloud-v2/controller.go
kevin85421: This is used to observe the behavior of "helm install" and "helm uninstall".
Without sleep, the helm chart will be uninstalled immediately after the helm chart installation, and thus we cannot observe the behavior of "helm install".

submarine-cloud-v2/pkg/helm/helm.go
kevin85421: I tested the function `os.Setenv` with the following example:
```go
 package main

 import (
     "fmt"
     "os"
 )

 func main() {
     os.Setenv("FOO", "1")
     fmt.Println("FOO:", os.Getenv("FOO"))
     fmt.Println("BAR:", os.Getenv("BAR"))
     fmt.Println("PATH:", os.Getenv("PATH"))
     fmt.Println()
 }
```

The result is shown in the following figure:
截圖 2021-04-13 下午1 54 56

According to the experiment result, the function `os.Setenv` will not influence other processes.

xunliu (approved):
LGMT.

PR #566, opened 2021-04-26
SUBMARINE-806. Create a submarine-database for single user

MortalHappiness:
### What is this PR for?

Create a submarine database for a single user via the submarine operator.

Reference: https://github.com/apache/submarine/blob/master/helm-charts/submarine/templates/submarine-database.yaml

### What type of PR is it?
[Feature]

### What is the Jira issue?

https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-806

```
# initialization
# Step1: Register Custom Resource Definition
kubectl apply -f artifacts/examples/crd.yaml

# Step2: Create a Custom Resource
kubectl apply -f artifacts/examples/example-submarine.yaml


# helm
cd submarine
helm install submarine ./helm-charts/submarine

# in-cluster
cd submarine/submarine-cloud-v2
eval $(minikube docker-env)
make image
kubectl apply -f artifacts/examples/submarine-operator-service-account.yaml
kubectl apply -f artifacts/examples/submarine-operator.yaml

# out-of-cluster
go build -o submarine-operator
./submarine-operator
```

+ PersistentVolume (helm)

helm-pv

+ PersistentVolume (operator)

**Note**: PersistentVolumes are not namespaced resources, so we add the namespace as a suffix to distinguish them

operator-pv

+ PersistentVolumeClaim (helm)

helm-pvc

+ PersistentVolumeClaim (operator)

operator-pvc

+ Deployment (helm)

helm-deployment

+ Deployment (operator)

operator-deployment

+ Service (helm)

helm-service

+ Service (opertator)

operator-service

### Questions:

### DIFF
--- a/submarine-cloud-v2/controller.go
+++ b/submarine-cloud-v2/controller.go
@@ -668,6 +668,42 @@ func (c *Controller) newSubmarineDatabase(namespace string, spec *v1alpha1.Subma
 
 	// TODO: (sample-controller) controller.go:287 ~ 293
 
+	// Step4: Create Service
+	service, service_err := c.serviceLister.Services(namespace).Get(databaseName)
+	// If the resource doesn't exist, we'll create it
+	if errors.IsNotFound(service_err) {
+		service, service_err = c.kubeclientset.CoreV1().Services(namespace).Create(context.TODO(),
+			&corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: databaseName,
+				},
+				Spec: corev1.ServiceSpec{
+					Ports: []corev1.ServicePort{
+						{
+							Port:       3306,
+							TargetPort: intstr.FromInt(3306),
+							Name:       databaseName,
+						},
+					},
+					Selector: map[string]string{
+						"app": databaseName,
+					},
+				},
+			},
+			metav1.CreateOptions{})
+		if deployment_err != nil {
+			klog.Info(service_err)
+		}
+		klog.Info("	Create Service: ", service.Name)
+	}
+	// If an error occurs during Get/Create, we'll requeue the item so we can
+	// attempt processing again later. This could have been caused by a
+	// temporary network failure, or any other transient reason.
+	if service_err != nil {
+		return service_err
+	}
+
+	// TODO: (sample-controller) controller.go:287 ~ 293
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


MortalHappiness:
@kevin85421 I've finished creating submarine-database for single user, thanks.

kevin85421:
@MortalHappiness Thank you for your contribution! Have you tested this PR with the in-cluster submarine operator?

MortalHappiness:
@kevin85421 Yes. I've tested it in both in-cluster ans out-cluster. The result is the same.

submarine-cloud-v2/controller.go
```diff
@@ -475,6 +483,230 @@ func (c *Controller) newSubmarineServerRBAC(serviceaccount_namespace string) err
...
+	// Step1: Create PersistentVolume
+	// PersistentVolumes are not namespaced resources, so we add the namespace
+	// as a suffix to distinguish them
+	pvName := databaseName + "-pv--" + namespace
+	pv, pv_err := c.persistentvolumeLister.Get(pvName)
+	// If the resource doesn't exist, we'll create it
+	if errors.IsNotFound(pv_err) {
+		var persistentVolumeSource corev1.PersistentVolumeSource
+		switch spec.Storage.StorageType {
+		case "nfs":
```
kevin85421, line 506: Does uppercase or lowercase matter in values.yaml?

MortalHappiness: Although in `submarine-database.yaml` there is a line `{{- if eq (.type | lower) "nfs" }}` so we can use "nfs" case-insensitively in our `values.yaml`, for example
```
storage:
    type: Nfs # "host" or "nfs"
    nfs:
      ip: 10.96.0.2
      path: "/"
```
as shown above, the value of *type* can be case-insensitive. However, the *nfs* variable below *type* still need to be case-sensitive. Therefore, I think it is unnecessary to support the case-insensitivity in only *type* parameter. I also write a regex validation in our `crd.yaml` to let the user know that the only valid values are "host" or "nfs".

https://github.com/apache/submarine/blob/2b3903e84468d28256cfa86979dfaff1aaebb32e/submarine-cloud-v2/artifacts/examples/crd.yaml#L76-L81

kevin85421 (approved):
LGTM

PR #584, opened 2021-05-10
SUBMARINE-825. Handle non-namespaced resources in delete event handler 

kevin85421:
#### Issue 1

Resources can be categorized into four categories

Type 1: namespaced resource created by client-go API

Type 2: non-namespaced resource created by client-go API, ex: persistent volume

Type 3: namespaced resource created by Helm API

Type 4: non-namespaced resource created by Helm API, ex: clusterrole (traefik)

Currently, the delete event resource handler can handle type 1, 2, 3, but this is not enough. Thus, this Jira aims to handle Type 4. If we do not handle Type 4, the error message will be shown as follows.

```
2021/05/11 09:58:50 rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "traefik" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "submarine-operator-test": current value is "submarine-operator-test-2"
```

#### Issue 2
Originally, we have two workers to handle the WorkQueueItems in the WorkQueue. However, it will cause some race conditions. To elaborate, worker A and worker B call `helm install traefik` at the same time, and thus one of the workers will report the error `release exist `.

To solve this problem, we change the number of workers to 1 temporarily. However, this solution is not good enough, and thus we need use concurrency techniques, such as Lock and Channel, to avoid these race conditions.

### What type of PR is it?
[Bug Fix]

### Todos
* Handle race conditions caused by multi workers
* Maintain `charts []helm.HelmUninstallInfo` for multi-tenant

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-825

```
# Step1: Out-of-cluster
go build -o submarine-operator
./submarine-operator

# Step2: Create submarine in namespace "submarine-operator-test"
kubectl create ns submarine-operator-test
kubectl apply -n submarine-operator-test -f artifacts/examples/example-submarine.yaml

# Step2: Delete Custom Resource
kubectl delete submarine example-submarine -n submarine-operator-test

# Step3: Check the result (non-namespaced resources)
kubectl get ns 
kubectl get pv
kubectl get clusterrole

# Step4: Create submarine in namespace "submarine-operator-test-2"
kubectl create ns submarine-operator-test-2
kubectl apply -n submarine-operator-test-2 -f artifacts/examples/example-submarine.yaml

# Step5: Delete Custom Resource
kubectl delete submarine example-submarine -n submarine-operator-test-2

# Step6: Check the result (non-namespaced resources)
kubectl get ns
kubectl get pv
kubectl get clusterrole
```

### Questions:

### DIFF
--- a/submarine-cloud-v2/README.md
+++ b/submarine-cloud-v2/README.md
@@ -62,10 +62,13 @@ make image
 kubectl apply -f artifacts/examples/submarine-operator-service-account.yaml
 
 # Step3: Deploy a submarine-operator
+kubectl apply -f artifacts/examples/submarine-operator.yaml
+
+# Step4: Deploy a submarine
 kubectl create ns submarine-operator-test
 kubectl apply -n submarine-operator-test -f artifacts/examples/example-submarine.yaml
 
-# Step4: Inspect submarine-operator POD logs 
+# Step5: Inspect submarine-operator POD logs 
 kubectl logs ${submarine-operator POD}
 ```
 
--- a/submarine-cloud-v2/controller.go
+++ b/submarine-cloud-v2/controller.go
@@ -93,6 +93,10 @@ type Controller struct {
 	// recorder is an event recorder for recording Event resources to the
 	// Kubernetes API.
 	recorder record.EventRecorder
+
+	// TODO: Need to be modified to implement multi-tenant
+	// Store charts
+	charts []helm.HelmUninstallInfo
 }
 
 const (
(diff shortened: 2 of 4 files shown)


kevin85421:
@Kenchu123 can you help me review this PR? Thanks!

Kenchu123 (approved):
LGTM.

PR #588, opened 2021-05-14
SUBMARINE-824. Execute port-forwarding automatically with Golang out-of-cluster

kevin85421:
### What is this PR for?
Execute the following command via Golang out-of-cluster
```
kubectl port-forward --address 0.0.0.0 -n ${Your_Namespace} service/traefik 32080:80
```

Reference: 
(1) https://gianarb.it/blog/programmatically-kube-port-forward-in-go
(2) https://github.com/minio/operator/blob/master/kubectl-minio/cmd/proxy.go

### What type of PR is it?
[Feature]

### Todos
* Execute port-forwarding automatically with Golang in-cluster
     *  The function `k8sutil.ServicePortForwardPort(context.TODO(), namespace, "traefik", 32080, 80, color.FgGreen)` does not work in-cluster.
     * (In submarine-operator pod) `curl 127.0.0.1:32080` --> Yes
     * (Local Chrome browser) 127.0.0.1:32080 --> 404
     * We need to forward the request from 127.0.0.1:32080 to the submarine-operator pod. (Use k8s NodePort to implement)

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-824

https://user-images.githubusercontent.com/20109646/118227081-86001080-b4ba-11eb-9274-6a053954f5f7.mov

### Questions:

### DIFF
--- a/submarine-cloud-v2/controller.go
+++ b/submarine-cloud-v2/controller.go
@@ -115,6 +115,7 @@ type WorkQueueItem struct {
 
 // NewController returns a new sample controller
 func NewController(
+	incluster bool,
 	kubeclientset kubernetes.Interface,
 	submarineclientset clientset.Interface,
 	traefikclientset traefik.Interface,
@@ -1142,8 +1143,11 @@ func (c *Controller) syncHandler(workqueueItem WorkQueueItem) error {
 		// TODO:
 		//   (1) multi-tenant port-forwarding
 		//   (2) Basic operations: on/off/modify (change port)
+		//   (3) in-cluster
 		if action == ADD {
-			k8sutil.ServicePortForwardPort(context.TODO(), namespace, "traefik", 32080, 80, color.FgGreen)
+			if !incluster {
+				k8sutil.ServicePortForwardPort(context.TODO(), namespace, "traefik", 32080, 80, color.FgGreen)
+			}
 		}
 	} else { // Case: DELETE
 		// Uninstall Helm charts
--- a/submarine-cloud-v2/controller.go
+++ b/submarine-cloud-v2/controller.go
@@ -21,14 +21,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
 	clientset "submarine-cloud-v2/pkg/generated/clientset/versioned"
 	submarinescheme "submarine-cloud-v2/pkg/generated/clientset/versioned/scheme"
 	informers "submarine-cloud-v2/pkg/generated/informers/externalversions/submarine/v1alpha1"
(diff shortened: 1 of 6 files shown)


kevin85421:
@Kenchu123 @MortalHappiness can you help me review this PR? Thanks!

submarine-cloud-v2/controller.go
```diff
@@ -112,6 +115,7 @@ type WorkQueueItem struct {
 
 // NewController returns a new sample controller
 func NewController(
+	incluster bool,
```
MortalHappiness, line 118: Here you pass `incluster` as parameter but not used.

submarine-cloud-v2/controller.go
```diff
@@ -1134,6 +1136,17 @@ func (c *Controller) syncHandler(workqueueItem WorkQueueItem) error {
...
 			return err
 		}
+
+		// Port-forwarding
+		// TODO:
+		//   (1) multi-tenant port-forwarding
+		//   (2) Basic operations: on/off/modify (change port)
+		//   (3) in-cluster
+		if action == ADD {
+			if !incluster {
```
MortalHappiness, line 1146: Here the `incluster` still works. But the `incluster` here is actually the global variable in `main.go`. I think what you want to do is to add this `incluster` as one attribute of `struct Controller` and use it.

submarine-cloud-v2/controller.go
```diff
@@ -1134,6 +1136,17 @@ func (c *Controller) syncHandler(workqueueItem WorkQueueItem) error {
...
+				k8sutil.ServicePortForwardPort(context.TODO(), newNamespace, "traefik", 32080, 80, color.FgGreen)
+			}
+		}
 	} else { // Case: DELETE
```
MortalHappiness, line 1150: You do not remove the port forward in delete event. So if the submarine custom resource is deleted and re-create again, some bug will happen.

MortalHappiness (changes requested):

submarine-cloud-v2/controller.go
kevin85421: I have fixed it.

submarine-cloud-v2/controller.go
kevin85421: I have fixed it.

kevin85421:
@MortalHappiness  Thank you for your recommendations. I have fixed the problems.

MortalHappiness (approved):
LGTM

Kenchu123 (approved):
LGTM

PR #591, opened 2021-05-25
SUBMARINE-830. Set OwnerReferences on resources

MortalHappiness:
### What is this PR for?
Set OwnerReferences on resources created using the spec of submarine custom resources so that:

1. When submarine custom resource is deleted, those resources will be deleted even when operator is not running
2. We can check whether the resources are created from the spec of submarine or not

References:

1. https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta
2. https://github.com/kubernetes/sample-controller/blob/73e81dab82c945b087f7db1f1f4a7b50e7d92751/controller.go#L388-L403
3. https://github.com/kubernetes/sample-controller/blob/73e81dab82c945b087f7db1f1f4a7b50e7d92751/controller.go#L289-L293

### What type of PR is it?
[Feature]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-830

1. Run the operator
2. Create a submarine custom resource
3. Stop the operator
4. Delete the submarine custom resource, and then all resoures created (except for resources created by helm) will be deleted

### Questions:

### DIFF
--- a/submarine-cloud-v2/controller.go
+++ b/submarine-cloud-v2/controller.go
@@ -68,6 +68,12 @@ import (
 
 const controllerAgentName = "submarine-controller"
 
+const (
+	// MessageResourceExists is the message used for Events when a resource
+	// fails to sync due to a Deployment already existing
+	MessageResourceExists = "Resource %q already exists and is not managed by Foo"
+)
+
 // Controller is the controller implementation for Foo resources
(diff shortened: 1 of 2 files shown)


kevin85421 (approved):
LGTM

PR #595, opened 2021-05-31
SUBMARINE-828. Set up event handlers for various resources other than submarine custom resource

MortalHappiness:
### What is this PR for?
Currently we only set up event handler for Submarine custom resource. We need to set up event handlers for other resources.

References:
1. https://github.com/kubernetes/sample-controller/blob/73e81dab82c945b087f7db1f1f4a7b50e7d92751/controller.go#L123-L142
2. https://github.com/kubernetes/sample-controller/blob/73e81dab82c945b087f7db1f1f4a7b50e7d92751/controller.go#L353-L386

### What type of PR is it?
[Feature]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-828

1. Run the operator using the command `./submarine-operator -v 4`
2. Create submarine custom resource
3. Delete the deployment `submarine-server`
4. See the log `Processing object: submarine-server` in the operator log

image

### Questions:

### DIFF
--- a/submarine-cloud-v2/controller.go
+++ b/submarine-cloud-v2/controller.go
@@ -188,7 +188,127 @@ func NewController(
 		},
 	})
 
-	// TODO: Setting up event handler for other resources. E.g. namespace
+	// Setting up event handler for other resources
+	namespaceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: controller.handleObject,
+		UpdateFunc: func(old, new interface{}) {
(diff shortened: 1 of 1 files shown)


MortalHappiness:
https://pkg.go.dev/k8s.io/apimachinery@v0.20.4/pkg/apis/meta/v1#OwnerReference

The documentation above says that we should not create cross-namespace ownerReferences but our submarine CR and the resources created by operator are in different namespaces. Therefore there are some bug in the `handleObject` function when we try to list the submarine CR in this PR. I will fix it later after discussion with others.

PR #606, opened 2021-06-13
SUBMARINE-848. Make submarine CR and resources created in the same namespace

MortalHappiness:
### What is this PR for?

https://pkg.go.dev/k8s.io/apimachinery@v0.20.4/pkg/apis/meta/v1#OwnerReference

The documentation above says that we should not create cross-namespace ownerReferences but our submarine CR and the resources created by operator are in different namespaces. After discussion with others, we decided to put them in the same namespace temporarily.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-848

### Questions:

### DIFF
--- a/submarine-cloud-v2/README.md
+++ b/submarine-cloud-v2/README.md
@@ -61,8 +61,9 @@ go build -o submarine-operator
 ./submarine-operator
 
 # Step2: Deploy a submarine
-kubectl create ns submarine-admin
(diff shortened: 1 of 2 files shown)


kevin85421 (approved):
LGTM

PR #617, opened 2021-06-29
SUBMARINE-871. Upgrade cglib for Ubuntu 20.04

KUAN-HSUN-LI:
### What is this PR for?
Running userSexDictAnnotationTest will add the DICT_SUFFIX in DictAnnotation class.
Using key with "@" will fail in cglib 3.2, but passing in cglib 3.3.
However, cglib is using asm 7.1, submarine-cluster and submarine-rpc are using asm 5.1. We should exclude the asm in both dependencies.

### What type of PR is it?
[Test]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-871

run the userSexDictAnnotationTest.

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
     <commons-configuration.version>1.10</commons-configuration.version>
     <commons-httpclient.version>3.1</commons-httpclient.version>
 
-    <cglib.version>3.2.2</cglib.version>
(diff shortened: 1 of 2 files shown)


ByronHsu:
LGTM.

PR #619, opened 2021-06-29
SUBMARINE-861. Trim the size of pysubmarine

featherchen:
### What is this PR for?

Trim the size of pysubmarine, in other words,we let the user install the package such as pytorch and Tensorflow by themself instead of pre-installation from pysubmarine
### What type of PR is it?
 Improvement 

### Todos
 [ ] - Deal with Jupiter Lab

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-861
### Questions:


featherchen:
@ByronHsu 
I'm not sure if I finished all the requirement, and  whether I should deal with the installation of package in Jupiter Lab or not.

ByronHsu:
Thanks for your contribution. 
I thought that we can rely on `extras_require` to get rid of explicitly installing ML framework.
`extras_require` means the dependencies for additional features, pip will not install them by default.
Users can use something like `pip install apache-submarine[feature1, feature2]` to explicitly install the dependencies for features.

ByronHsu:
@featherchen Yes, you should modify the jupyter dockerfile too.

featherchen:
@ByronHsu

featherchen:
@ByronHsu  I failed on the Integration test,maybe it is the bug you mentioned earlier?

featherchen:
@ByronHsu  I wonder why in integration test it didn't pip install pytorch and Tensorflow,which are be wriiten in ./github/workflow/python.yaml.An alternative solution is that use pip install --no-cache-dir ./submarine-sdk/pysubmarine/.[tf.pytorch] , but it didn't really fix the problem though.
Screenshot from 2021-07-03 01-09-32

ByronHsu:
@featherchen There are two workflows in python.yml: unit test and integration test. Only unit test install pytorch and tf beforehand, but integration did not. You can check here: https://github.com/apache/submarine/blob/master/.github/workflows/python.yml#L133.
I wonder why `pip install --no-cache-dir ./submarine-sdk/pysubmarine/.[tf.pytorch]` cannot solve the problem?

featherchen:
> @featherchen There are two workflows in python.yml: unit test and integration test. Only unit test install pytorch and tf beforehand, but integration did not. You can check here: https://github.com/apache/submarine/blob/master/.github/workflows/python.yml#L133.
> I wonder why `pip install --no-cache-dir ./submarine-sdk/pysubmarine/.[tf.pytorch]` cannot solve the problem?

ByronHsu:
Thanks for your contribution! LGTM.
Can you open an issue to update the usage of submarine-sdk?

PR #620, opened 2021-06-29
SUBMARINE-874. Use the locally installed kind and kubectl

KUAN-HSUN-LI:
### What is this PR for?
This PR is a preparation of the integration test in GitHub Actions. Since there is the wrong version of kubectl and KinD installed in GitHub Actions, we should specify our own versions locally.

### What type of PR is it?
[Test]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-874

See the test-k8s in Travis CI.

### Questions:

### DIFF
--- a/submarine-cloud/hack/lib.sh
+++ b/submarine-cloud/hack/lib.sh
@@ -47,16 +47,18 @@ function hack::ensure_kubectl() {
         return 0
     fi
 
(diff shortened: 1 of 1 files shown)


ByronHsu:
LGTM. But submarine-cloud might be deprecated soon.

PR #622, opened 2021-06-29
SUBMARINE-880. Cannot run two experiments with the same name

kevin85421:
### What is this PR for?
We cannot run two experiments with the same name. Take experimentIT.java as an example, the frontend E2E testcase will create an experiment named "experiment-e2e-test". If we run this testcase twice, the workbench just shows an "experiment-e2e-test" rather than two.

The root cause is that the K8sSubmitter will create two PODs for an "experiment-e2e-test". The names of these two PODs are "experiment-e2e-test-ps-0" and "experiment-e2e-test-worker-0". Hence, when we try to create the second "experiment-e2e-test", K8sSubmitter will throw an exception due to duplicate POD names.

To reproduce the bug:
```bash
# Step1: Run workbench on port 8080
# Step2: 
cd submarine-cloud-v2

# Step3: Create "experiment-e2e-test" twice
./hack/run_frontend_e2e.sh experimentIT
./hack/run_frontend_e2e.sh experimentIT

# Step4: Check Workbench: only an "experiment-e2e-test" exists => BUG!
```

* My solution: (Outdated)
  * The variable "name" indicates the name of Custom Resource (ex: tfjobs, pytorchjobs) in k8s.
  * The variable "expName" indicates the name of experiment.
  * ${name} = ${expName} + "-" + ${experimentCounter}
  * For example: an user creates an experiment "experiment-e2e-test"  (${expName})
    * On the workbench, the experiment name is "experiment-e2e-test". (${expName})
    * The name of the tfjobs (a CRD) is "experiment-e2e-test-0001". (${name})
    * The value of the label "submarine-experiment-name" in the TFJob is "experiment-e2e-test"  (${expName})
    * The names of PODs created by the TFJob "experiment-e2e-test-0001" are: 
      ```
      experiment-e2e-test-0001-ps-0
      experiment-e2e-test-0001-worker-0
      ```

### What type of PR is it?
[Bug Fix]

### Todos
* Check "_" (Reference: [Link](https://blog.csdn.net/airangrong6572/article/details/101273245))

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-880

```bash
# Step1: Run workbench on port 8080
# Step2: 
cd submarine-cloud-v2

# Step3: Create "experiment-e2e-test" twice
./hack/run_frontend_e2e.sh experimentIT
./hack/run_frontend_e2e.sh experimentIT

# Step4: Check Workbench: two"experiment-e2e-test" exists (Correct)
```

截圖 2021-07-08 上午1 58 10
截圖 2021-07-08 上午1 59 43
截圖 2021-07-08 上午1 57 49
截圖 2021-07-08 上午1 58 00

### Questions:

### DIFF
--- a/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
+++ b/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
@@ -28,8 +28,8 @@ public class ExperimentMeta {
 
   public static final String SUBMARINE_EXPERIMENT_NAME = "submarine-experiment-name";
 
+  private String experimentId;
   private String name;
-  private String expName;
   private String namespace;
   private String framework;
   private String cmd;
@@ -56,19 +56,19 @@ public class ExperimentMeta {
   }
 
   /**
-   * Get the exp name which is unique within a namespace.
-   * @return exp name
+   * Get the experiment id which is unique within a namespace.
+   * @return experiment id
    */
-  public String getExpName() {
-    return expName;
+  public String getExpID() {
+    return experimentId;
   }
 
   /**
-   * Exp Name must be unique within a namespace. Is required when creating experiment.
-   * @param name exp name
+   * experiment id must be unique within a namespace. Is required when creating experiment.
+   * @param name experiment name
    */
-  public void setExpName(String expName) {
-    this.expName = expName;
+  public void setExpID(String experimentId) {
+    this.experimentId = experimentId;
   }
 
   /**
--- a/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
+++ b/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
@@ -29,7 +29,7 @@ public class ExperimentMeta {
   public static final String SUBMARINE_EXPERIMENT_NAME = "submarine-experiment-name";
 
   private String name;
-  private String podName;
+  private String expName;
   private String namespace;
   private String framework;
   private String cmd;
@@ -56,19 +56,19 @@ public class ExperimentMeta {
   }
 
   /**
-   * Get the pod name which is unique within a namespace.
-   * @return pod name
+   * Get the exp name which is unique within a namespace.
+   * @return exp name
    */
-  public String getPodName() {
-    return podName;
+  public String getExpName() {
+    return expName;
   }
 
   /**
-   * POD Name must be unique within a namespace. Is required when creating experiment.
-   * @param name pod name
+   * Exp Name must be unique within a namespace. Is required when creating experiment.
+   * @param name exp name
    */
-  public void setPodName(String podName) {
-    this.podName = podName;
+  public void setExpName(String expName) {
+    this.expName = expName;
   }
 
   /**
--- a/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
+++ b/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
@@ -59,7 +59,7 @@ public class ExperimentMeta {
    * Get the experiment id which is unique within a namespace.
    * @return experiment id
    */
-  public String getExpID() {
+  public String getExperimentId() {
     return experimentId;
   }
 
@@ -67,7 +67,7 @@ public class ExperimentMeta {
    * experiment id must be unique within a namespace. Is required when creating experiment.
    * @param experimentId experiment id
    */
-  public void setExpID(String experimentId) {
+  public void setExperimentId(String experimentId) {
     this.experimentId = experimentId;
   }
 
--- a/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
+++ b/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
@@ -25,7 +25,11 @@ import java.util.Map;
  * ExperimentMeta is metadata that all experiments must have.
  */
 public class ExperimentMeta {
+
+  public static final String SUBMARINE_EXPERIMENT_NAME = "submarine-experiment-name";
+
   private String name;
+  private String podName;
   private String namespace;
   private String framework;
   private String cmd;
@@ -49,6 +53,22 @@ public class ExperimentMeta {
    */
   public void setName(String name) {
     this.name = name;
+  }
+
+  /**
+   * Get the pod name which is unique within a namespace.
+   * @return pod name
+   */
+  public String getPodName() {
+    return podName;
+  }
+
+  /**
+   * POD Name must be unique within a namespace. Is required when creating experiment.
+   * @param name pod name
+   */
+  public void setPodName(String podName) {
+    this.podName = podName;
   }
 
   /**
--- a/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
+++ b/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
@@ -65,7 +65,7 @@ public class ExperimentMeta {
 
   /**
    * experiment id must be unique within a namespace. Is required when creating experiment.
-   * @param name experiment name
+   * @param experimentId experiment id
    */
   public void setExpID(String experimentId) {
     this.experimentId = experimentId;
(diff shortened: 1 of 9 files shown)


jiwq (commented):
I'm not sure whether need to append the counter after the job name.

As I know everyone has an independent server instance and run in different namespace, so  I don't recommend intro this enhancement. If missing something, please correct me. :)

kevin85421:
@jiwq Thank you for your recommendation! In my opinion, data scientists need to run multiple experiments to achieve a single task. 

For example, a data scientist wants to train a model with the MNIST dataset. At first, he creates an experiment "mnist-exp" with his `minist-ver1.py`. Next, because the model accuracy does not satisfy the scientist's expectation, the data scientist makes some modifications to the Python script (`mnist-ver1.py`) and creates an experiment "mnist-exp" again. The bug will occur seamlessly (That is, the data scientist does not know his experiment will not be created).

To avoid the condition, the data scientist needs to:

* Method1: 
  * Delete the first "mnist-exp"
  * Create "mnist-exp" with the updated `mnist-ver1.py`.
  * Cons: 
    * (1) Lose the data about the first "mnist-exp" 
    * (2) Delete "mnist-exp" manually
* Method2:
  * Change the name of the experiment (ex: "mnist-exp-2")
  * Cons: Data scientists usually need to run many experiments to achieve a single task. Thus, the repeat human effort is very annoying. (For example, "mnist-exp-2", "mnist-exp-3", "mnist-exp-4" ... "mnist-exp-100")


With my personal user experience, I usually use a script to help me create an experiment without entering the information, including experiment name, number of workers, and resource limitations, again and again. However, because of this bug, when I want to run my script to create an experiment again, I need a lot of human effort.

ByronHsu:
I agree with @kevin85421

jiwq:
Thanks @kevin85421 for a detailed explanation. I agree, but I think we should set the job name as the pod label. So that the user can retrieve his/her experiments easily by name.

In this patch we should set the pod label. I accept add the new api later to retrieve the experiments by the user's experiment name, and we should fill a JIRA to track this feature. Any thoughts?

kevin85421:
@jiwq Thank you for your recommendations! 

I have some questions:
Q1: Does "job name" mean "experiment name"?
Q2: Does "pod label" mean "pod name"?

The following points assumed that the answers to these two questions are YES. In my opinion, with this patch, users can still retrieve their experiments via experiment names easily because the experiment (with the modified name) will be returned. Hence, people can get the experiment easily from the workbench.

In addition, because the rename rule is appending ${experimentCounter} to the original experiment name, users can still retrieve their experiment with the original experiment name easily.

For example, a user creates an experiment "mnist-exp".

Experiment name: "mnist-exp-0001" (same as the experiment name on the workbench)
POD name: "mnist-exp-0001"

When a user wants to retrieve the experiment, they can use both "mnist-exp-0001" or "mnist-exp".

jiwq:
> Q1: Does "job name" mean "experiment name"?

Yes, it is experiment spec name field. 

Because we defined the name in experiment spec, so I think we should keep this concept and user should not to know the real pod name. 

> Q2: Does "pod label" mean "pod name"?

NO. More info see https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/  

By example, user defined the mnist tensorflow experiment spec in submarine and submit twice, we can found mnist-1 and mnist-2 jobs. If user want to query the minist experiment, he/her only should use the mnist as the query param to retrieve the list of experiments. And if want get more info should use the specific name we can call it instance name, which is mnist-1 or mnist-2 etc.

Simplify, I think we should two tasks to do.
1.  Fill the K8s object labels use experiment name field.
2. Append the counter after the job(tfjob/pytorch job)'s name which is defined in experiment spec.

kevin85421:
@jiwq Got it. It makes sense.

To conclude, 
(1) The experiment name on the workbench must be the same as the name specified by users. 
(2) The names of the PODs (i.e. instance name) related to the experiment are `${original_exp_name} + "-" + ${experimentCounter}`
(3) Users can use ${original_exp_name} to retrieve their experiments.
(4) Set the labels of the PODs to ${original_exp_name}

For instance,

User wants to create an experiment: ${original_exp_name} = "mnist-exp"
The experiment name on the workbench: "mnist-exp" => conclusion (1)
Names of PODs: "mnist-exp-0001" => conclusion (2)
Labels of PODs: "mnist-exp" => conclusion (4)

Do I misunderstand anything? Thanks!

jiwq:
@kevin85421 Perfect :) I'm not sure whether the workbench supports the feature. If not we should develop.

kevin85421:
@jiwq I have updated this PR. You can check the sections: "My solution" and "Screenshots".

To elaborate, 

an user creates an experiment: ${original_exp_name} = "experiment-e2e-test"
The experiment name on the workbench: "experiment-e2e-test" => conclusion (1)
Names of PODs: "experiment-e2e-test-0001" => conclusion (2)
Name of CR: "experiment-e2e-test-0001"
Labels of CR: "experiment-e2e-test" => conclusion (4)

jiwq:
@kevin85421 How about use the experiment id as the pod name?

kevin85421:
@jiwq I have fixed it. Thanks!

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
```diff
@@ -51,6 +55,22 @@ public void setName(String name) {
     this.name = name;
   }
 
+  /**
+   * Get the experiment id which is unique within a namespace.
+   * @return experiment id
+   */
+  public String getExpID() {
```
jiwq, line 62: Disallow to abridge the name, because it will reduce readability.
```suggestion
  public String getExperimentId() {
```

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
```diff
@@ -51,6 +55,22 @@ public void setName(String name) {
...
+    return experimentId;
+  }
+
+  /**
+   * experiment id must be unique within a namespace. Is required when creating experiment.
+   * @param experimentId experiment id
+   */
+  public void setExpID(String experimentId) {
```
jiwq, line 70:
```suggestion
  public void setExperimentId(String experimentId) {
```

submarine-server/server-core/src/main/java/org/apache/submarine/server/experiment/ExperimentManager.java
```diff
@@ -108,8 +108,10 @@ public Experiment createExperiment(ExperimentSpec spec) throws SubmarineRuntimeE
     spec.getMeta().getEnvVars().put(RestConstants.SUBMARINE_TRACKING_URI, url);
     spec.getMeta().getEnvVars().put(RestConstants.LOG_DIR_KEY, RestConstants.LOG_DIR_VALUE);
 
-    String lowerName = spec.getMeta().getName().toLowerCase();
+    String lowerName = spec.getMeta().getName().toLowerCase(); 
     spec.getMeta().setName(lowerName);
+    spec.getMeta().setExpID(id.toString().replaceAll("_", "-"));
```
jiwq, line 113: What's the reason of considering '-' rather than '_'? 

Supported character set of K8s: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

kevin85421: #630 describes that why I replace "_" with "-".

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
kevin85421: I have fixed it.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentMeta.java
kevin85421: I have fixed it.

kevin85421:
@jiwq I am sorry to bother you. Do you have any other recommendations about this patch? Thanks!

submarine-server/server-core/src/main/java/org/apache/submarine/server/experiment/ExperimentManager.java
jiwq: Yes. But the label value could contain dashes (-), underscores (_), dots (.), and alphanumerics between. 

I think we should keep it, so that we can't convert the concept when use label selector. Any thoughts?

jiwq: +1 After discussed offline. Thanks @kevin85421 for the future design, I expected the refactor of experiment feature.

kevin85421:
Thank @jiwq for your recommendations! I will propose a new design to enable experiments to be grouped by tags.

PR #626, opened 2021-06-30
SUBMARINE-829. Handle errors when processing workqueue items

MortalHappiness:
### What is this PR for?
Currently we do not handle error when processing workqueue items, we need to handle errors when processing workqueue items.

Reference: https://github.com/kubernetes/sample-controller/blob/73e81dab82c945b087f7db1f1f4a7b50e7d92751/controller.go#L195-L230

### What type of PR is it?
[Feature]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-829

### Questions:


kevin85421 (approved):
LGTM

PR #627, opened 2021-06-30
SUBMARINE-882. fix datadictIT

noidname01:
### What is this PR for?
In travis.ci, this IT will have undeterministic behavior, it is because of the original way to realize the waiting is fluent wait, but it lacks of expected condition.
The other reason of unstable behavior of this test is because of the wrong interaction with components.

https://user-images.githubusercontent.com/55401762/124241557-6e561780-db4e-11eb-8ab4-53a04ffc583e.mp4

As you can see, this drop-down list need to be hovered. But this IT clicked it. This might be the source of unstable.
So it need to be modify with the functions that use explicit wait in AbstractSubmarineIT.
Also found that the window need to set larger to prevent element from disappearing in viewpoint, which makes headless driver can not interact.
Note: the default of size of headless window maximize is (600, 800), when driver hover the target out of this area, it will cause MoveTargetOutOfBounds exception.

### What type of PR is it?
[Bug Fix | Refactoring]

### What is the Jira issue?

https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-882

```bash
# Step1: Run Submarine workbench on 127.0.0.1:8080 by operator
# Step2: Run testcase
cd submarine-cloud-v2

# Usage: ./hack/run_frontend_e2e.sh ${testcase}
# Example: 
./hack/run_frontend_e2e.sh datadictIT
```

https://user-images.githubusercontent.com/55401762/124000807-8c186500-da06-11eb-8289-faf054a85650.mp4

### Questions:

### DIFF
--- a/submarine-test/test-e2e/src/test/java/org/apache/submarine/AbstractSubmarineIT.java
+++ b/submarine-test/test-e2e/src/test/java/org/apache/submarine/AbstractSubmarineIT.java
@@ -37,6 +37,7 @@ import org.openqa.selenium.support.ui.Wait;
 import org.openqa.selenium.support.ui.WebDriverWait;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.openqa.selenium.interactions.Actions;
 
 import java.io.BufferedReader;
 import java.io.File;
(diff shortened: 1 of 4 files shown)


kevin85421 (commented):
@noidname01 Thank you for your contribution!

There are some resources that can help you figure out why does non-deterministic behavior happen.

(1) You need to understand the difference between implicit wait, explicit wait, and fluent wait.
https://www.selenium.dev/documentation/en/webdriver/waits/

(2) Try to understand PR 605
https://github.com/apache/submarine/pull/605/files

With these two resources, you can fix the error on Travis, and make your code more concise.

kevin85421 (approved):
LGTM. Good job @noidname01

PR #637, opened 2021-07-03
SUBMARINE-889. Change module import path for controller

MortalHappiness:
### What is this PR for?

Change module import path from `submarine-cloud-v2` to `github.com/apache/submarine/submarine-cloud-v2` so that our packages can be viewed on https://pkg.go.dev

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-889

This is a sub-task of https://issues.apache.org/jira/browse/SUBMARINE-888

### Questions:


kevin85421 (approved):
LGTM

PR #639, opened 2021-07-03
SUBMARINE-890. Extract all resource creation spec

MortalHappiness:
### What is this PR for?
Extract all resource creation spec into functions so the code is more readable.

### What type of PR is it?
[Refactoring]

### Todos
* [x] - Make controller be a standalone package
* [x] - Extract resource creation spec into functions

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-890

### Questions:


MortalHappiness:
@kevin85421 Can you help me review this PR? Thank you.

kevin85421 (approved):
LGTM

PR #653, opened 2021-07-07
SUBMARINE-909. Add API to get the serve pod's status.

jeff-901:
### What is this PR for?
Add http API to get the serve pod's ready status.

### What type of PR is it?
Improvement

### Todos
* [ ] - Add serve id

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-909

Screenshot from 2021-07-07 17-23-29

### Questions:


submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/K8sSubmitter.java
```diff
@@ -534,6 +535,49 @@ public ServeResponse deleteServe(ServeRequest spec)
...
+        if (item.getMetadata().getLabels().get("app").equals(name)){
+          pod = item;
+          break;
+        }
+      } catch (NullPointerException e){
+        continue;
+      }
+    }
+    if (pod == null) {
+      throw new SubmarineRuntimeException(404, "Pod " + name + " not fund.");
```
ByronHsu, line 566: The error code should be replaced with constant

jeff-901: But some functions above is not using constant error code.

PR #674, opened 2021-07-13
SUBMARINE-924. [User] Python SDK Model Client

featherchen:
### What is this PR for?

Documentation of Python SDK Model Client for users.
### What type of PR is it?
Documentation

### Todos
* [ ] - Test load_model , update_model, delete_model 

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-924

### Questions:

### DIFF
--- a/website/docs/userDocs/submarine-sdk/model-client.md
+++ b/website/docs/userDocs/submarine-sdk/model-client.md
@@ -20,3 +20,123 @@ KIND, either express or implied.  See the License for the
 specific language governing permissions and limitations
 under the License.
 -->
+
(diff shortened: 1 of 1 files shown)


featherchen:
@ByronHsu Is there an easier way to test the sdk I mention above ?
The way came to my mind is that post another request of .py to connect to the server, but this way is quite complicated that needs to post two time (one for training, one for update/load/delete).

ByronHsu:
@featherchen You can use telepresence to connect to mlflow server on local. It might be faster to test on this way.

ByronHsu:
LGTM.

PR #686, opened 2021-07-19
SUBMARINE-944. Bump io:commons-io to 2.11

pingsutw:
### What is this PR for?

CVE-2021-29425
Vulnerable versions: < 2.7
Patched version: 2.7
In Apache Commons IO before 2.7, When invoking the method FileNameUtils.normalize with an improper input string, like "//../foo", or "\..\foo", the result would be the same value, thus possibly providing access to files in the parent directory, but not further above (thus "limited" path traversal), if the calling code would use the result to construct a path value.
https://github.com/apache/submarine/security/dependabot/pom.xml/commons-io:commons-io/open 

### What type of PR is it?
[Improvement]

### Todos
No

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-944

Pass the CIs
### Questions:

### DIFF
--- a/LICENSE-binary
+++ b/LICENSE-binary
@@ -232,7 +232,7 @@ commons-configuration:commons-configuration:1.6
 commons-configuration:commons-configuration:1.1
 commons-daemon:commons-daemon:1.0.13
 commons-digester:commons-digester:1.8
-commons-io:commons-io:2.4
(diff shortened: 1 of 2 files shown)


kevin85421 (approved):
LGTM

PR #689, opened 2021-07-23
SUBMARINE-911. Fix the bug about deleting notebook

aeioulisa:
### What is this PR for?

Store notebook data to database.
Those data used to be saved in cache. The notebook pod have persistent volume to retain them which make the pod still working even if the server restart for some reason. But in this case, we will lose control of the pod on workbench because we don't have the data of these notebooks.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-911

image

### Questions:

### DIFF
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
@@ -153,7 +153,7 @@ public class NotebookManager {
     Notebook notebook = notebookService.select(id);
     if (notebook == null) {
       throw new SubmarineRuntimeException(Response.Status.NOT_FOUND.getStatusCode(),
-          "Not found notebook");
+          "Notebook not found.");
     }
     Notebook foundNotebook = submitter.findNotebook(notebook.getSpec());
     foundNotebook.rebuild(notebook);
@@ -203,7 +203,7 @@ public class NotebookManager {
     NotebookId notebookId = NotebookId.fromString(id);
     if (notebookId == null) {
       throw new SubmarineRuntimeException(Response.Status.NOT_FOUND.getStatusCode(),
-          "Not found notebook server.");
+          "Notebook not found.");
     }
   }
 
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
@@ -32,39 +32,37 @@ import org.apache.submarine.server.environment.EnvironmentManager;
 import javax.ws.rs.core.Response;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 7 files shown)


submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
```diff
@@ -118,79 +118,89 @@ public Notebook createNotebook(NotebookSpec spec) throws SubmarineRuntimeExcepti
...
       throw new SubmarineRuntimeException(Response.Status.OK.getStatusCode(),
-              "Invalid. Notebook Spec object is null.");
+          "Invalid. Notebook Spec object is null.");
     }
   }
 
   private void checkNotebookId(String id) throws SubmarineRuntimeException {
     NotebookId notebookId = NotebookId.fromString(id);
-    if (notebookId == null || !cachedNotebookMap.containsKey(id)) {
+    if (notebookId == null) {
```
pingsutw, line 200: Don't we need to check if notebookId is in the database?

submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
```diff
@@ -118,79 +118,93 @@ public Notebook createNotebook(NotebookSpec spec) throws SubmarineRuntimeExcepti
...
-    Notebook notebook = cachedNotebookMap.get(id);
-    NotebookSpec spec = notebook.getSpec();
-    Notebook patchNotebook = submitter.findNotebook(spec);
-    notebook.rebuild(patchNotebook);
-    return notebook;
+
+    Notebook notebook = notebookService.select(id);
+    if (notebook == null) {
+      throw new SubmarineRuntimeException(Response.Status.NOT_FOUND.getStatusCode(),
+          "Not found notebook");
```
pingsutw, line 156:
```suggestion
          "Notebook not found.");
```

submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
```diff
@@ -118,79 +118,93 @@ public Notebook createNotebook(NotebookSpec spec) throws SubmarineRuntimeExcepti
...
     }
   }
 
   private void checkNotebookId(String id) throws SubmarineRuntimeException {
     NotebookId notebookId = NotebookId.fromString(id);
-    if (notebookId == null || !cachedNotebookMap.containsKey(id)) {
+    if (notebookId == null) {
       throw new SubmarineRuntimeException(Response.Status.NOT_FOUND.getStatusCode(),
-              "Not found notebook server.");
+          "Not found notebook server.");
```
pingsutw, line 206:
```suggestion
          "Notebook not found.");
```

pingsutw (approved):
other LGTM

PR #694, opened 2021-08-03
SUBMARINE-952. add backoffLimit and Failed status for experiment object

FatalLin:
### What is this PR for?
just like we mentioned in Jira ticket, for now submarine will retry those retry able jobs endlessly even those job never had a chance to success. It's waste of resource obviously, so I add a MLJob property BackoffLimit to prevent this kind of situation, at same time I change the MLJobSpec from interface into abstract class to share property with TFJobSpec and PytorchJobSpec.
I also fixed a bug to respond the correct status of experiment in failure case. 

### What type of PR is it?
Improvement

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-952

modify the test case (https://github.com/apache/submarine/blob/master/submarine-test/test-e2e/src/test/java/org/apache/submarine/integration/experimentIT.java#L90) from {1024, 1024} to {512, 512},
and the experiment will hit OOMFailure, and the experiment status will change into failed after retry 3 times.
截圖 2021-08-01 下午5 03 43
截圖 2021-08-03 下午11 10 39

### Questions:

### DIFF
--- a/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/experiment/Experiment.java
+++ b/submarine-server/server-api/src/main/java/org/apache/submarine/server/api/experiment/Experiment.java
@@ -123,7 +123,8 @@ public class Experiment {
     STATUS_CREATED("Created"),
     STATUS_RUNNING("Running"),
     STATUS_SUCCEEDED("Succeeded"),
-    STATUS_DELETED("Deleted");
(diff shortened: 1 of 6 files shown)


pingsutw (approved):
LGTM

PR #695, opened 2021-08-04
SUBMARINE-962. Delete temporary build artifacts before caching

pingsutw:
### What is this PR for?

When building the Submarine repo, some of the modules would use outdated packages in `~/.m2/repository/org/apache` in Github action cache
We should delete temporary build artifacts before caching
I believe it will fix the test failures

### What type of PR is it?
[Bug Fix]

### Todos
No

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-962

Pass the CIs
### Questions:


pingsutw:
cc @aeioulisa Could you help me review it?

PR #697, opened 2021-08-05
SUBMARINE-953. Submarine-server shouldn't create persistentVolume manually for notebook

Kenchu123:
### What is this PR for?

Source: https://github.com/apache/submarine/blob/72be805fc4672f7f29c96cca95e16d59358d4dd2/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/K8sSubmitter.java#L379

We can find out that when we create a notebook custom resource, the server will manually create the persistent volume for it.

However, this made submarine-server has to grant access to persistent volume resources, which lead to setting cluster roles for the server. (https://github.com/apache/submarine/blob/72be805fc4/helm-charts/submarine/templates/rbac.yaml#L61)

Since the server is in namespace scope, giving it cluster roles is inappropriate. Besides, this makes multi-tenancy more difficult.

To fix the bug, we can create a storage class to dynamically provision persistent volumes, and submarine-server will only need to deal with persistent volume claim.

Ref:
- storage class: https://kubernetes.io/docs/concepts/storage/storage-classes/#local
- notebook-cr spec: https://www.kubeflow.org/docs/reference/notebook/v1/ 

### What type of PR is it?
[Bug Fix]

### Todos

* [x] - add storageClass
* [x] - add persistentVolumeClaim spec to submarine-server
* [x] - remove submarine-server createPersistentVolume
* [x] - replace clutserrolebinding to rolebinding in helm-charts
* [x] - replace clusterrole to role in helm-charts
* [x] - modify submarine operator 

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-953

Can be tested either using `helm install` or `submarine-operator`.

https://user-images.githubusercontent.com/17617373/128290829-ff38828c-3555-4c0c-b37d-6b147b762dde.mov

### Questions:

### DIFF
--- a/helm-charts/submarine/charts/notebook-controller/templates/storageclass.yaml
+++ b/helm-charts/submarine/charts/notebook-controller/templates/storageclass.yaml
@@ -1,7 +1,7 @@
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
-  name: "notebook-storageclass"
+  name: notebook-storageclass
   labels:
     app: notebook-controller
 provisioner: k8s.io/minikube-hostpath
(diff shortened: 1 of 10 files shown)


pingsutw (approved):
LGTM

PR #621, opened 2021-06-29
SUBMARINE-856. Predefined experiment pods CrashLoopBackOff

aeioulisa:
### What is this PR for?
Solve the problem of predefined experiment pods CrashLoopBackOff.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-856

### Questions:
* Do the license files need updating? Yes/No
* Are there breaking changes for older versions? Yes/No
* Does this need new documentation? Yes/No

PR #510, opened 2021-02-11
SUBMARINE-728. The experiment example will cause OOM error.

kevin85421:
### What is this PR for?
The experiment example will cause OOM error.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-728?filter=myopenissues

### How should this be tested?
* Step1: Create an experiment
```
curl -X POST -H "Content-Type: application/json" -d '
{
  "meta": {
    "name": "tf-mnist-json",
    "namespace": "default",
    "framework": "TensorFlow",
    "cmd": "python /var/tf_mnist/mnist_with_summaries.py --log_dir=/train/log --learning_rate=0.01 --batch_size=150",
    "envVars": {
      "ENV_1": "ENV1"
    }
  },
  "environment": {
    "image": "apache/submarine:tf-mnist-with-summaries-1.0"
  },
  "spec": {
    "Ps": {
      "replicas": 1,
      "resources": "cpu=1,memory=1024M"
    },
    "Worker": {
      "replicas": 1,
      "resources": "cpu=1,memory=2048M"
    }
  }
}
' http://127.0.0.1:8080/api/v1/experiment
```
* Step2: Check the status of the Pods
```
kubectl get pods
```
### Questions:
* Does the licenses files need update? No
* Is there breaking changes for older versions? No
* Does this needs documentation? No

PR #699, opened 2021-08-06
SUBMARINE-948. Allow experiments to overcommit memory

kevin85421:
### What is this PR for?
The following two pull requests aim to resolve the Out-Of-Memory error. However, it is very inconvenient for users to predict the actual memory usage. Thus, using the memory request and memory limit mechanism to allow overcommitment of memory is helpful for users.

* https://github.com/apache/submarine/pull/621
* https://github.com/apache/submarine/pull/510

In this PR, I set the memory limit to twice the memory request to enable overcommitment of memory. With this patch, the OOM errors can be reduced effectively.

This [article](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-resource-requests-and-limits) is a good resource to better understand this PR.

### What type of PR is it?
[Feature]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-948

**Test1**
* Create a distributed TensorFlow MNIST job, and set the memory quota of a worker to 512 MB. To elaborate, modify [experimentIT.java:90](https://github.com/apache/submarine/blob/master/submarine-test/test-e2e/src/test/java/org/apache/submarine/integration/experimentIT.java#L90) to 
  ```java
  experimentPage.fillTfSpec(2, new String[]{"Ps", "Worker"}, new int[]{1, 1}, new int[]{1, 1}, new int[]{512, 512});
  ```
* Without this PR, this MNIST job will be killed due to an Out-Of-Memory error. On the other hand, with this PR, the MNIST job will not be killed.

**Test2**
* The memory limit is equal to twice the memory request.
    ```
    kubectl describe ${your_experiment_pod}
    ```
    截圖 2021-08-06 下午2 40 42

* Kubernetes integration test on my local machine
截圖 2021-08-06 下午3 53 50

### Questions:

### DIFF
--- a/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
+++ b/submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/parser/ExperimentSpecParser.java
@@ -61,9 +61,7 @@ import java.util.List;
 import java.util.Map;
 
 public class ExperimentSpecParser {
-
-  private static SubmarineConfiguration conf =
-      SubmarineConfiguration.getInstance();
+  private static SubmarineConfiguration conf = SubmarineConfiguration.getInstance();
 
   public static MLJob parseJob(ExperimentSpec experimentSpec) throws InvalidSpecException {
     String framework = experimentSpec.getMeta().getFramework();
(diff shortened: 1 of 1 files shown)


kevin85421:
@pingsutw Can you help me review this PR? Thanks!

pingsutw (approved):
LGTM

PR #683, opened 2021-07-17
SUBMARINE-942. Make experiment ID consistent with TFJob and PyTorch Job

kevin85421:
### What is this PR for?
1. Please refer to the following two JIRA issues.

   * https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-886?filter=reportedbyme
   * https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-880?filter=reportedbyme

   In this JIRA issue, we need to make the experiment ID consistent with the name of TFJob and PyTorch Job. The difference is caused by this [link](https://github.com/apache/submarine/commit/8da9f478de9323dd098d06bbd52afdda3a27ce07#diff-c1cad0951ad3663eb075d964171d74ae94a35bec2b79f8c9894030c0446a6b99R119). Update ExperimentId.java.

2. Update [IntegrationTestK8s.md](http://submarine.apache.org/docs/devDocs/IntegrationTestK8s)
    * "submarine-server-core" is a dependency package specified in pom.xml of test-k8s.
    * In the document [BuildFromCode.md](http://submarine.apache.org/docs/devDocs/BuildFromCode), the command to build Submarine is `mvn clean package -DskipTests`. However, the package will be installed into the local repository at **install phase**, a later phase than both **package and verify phases**.
    * Hence, we need to execute `mvn install -DskipTests` to ensure that the **test-k8s** module uses the latest "submarine-server-core" module.
 
3. Remove the field `name` in Experiment.java because the value of `name` is the same as `experimentId`.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-942

### Questions:
* Do the license files need updating? No
* Are there breaking changes for older versions? No
* Does this need new documentation? No

PR #702, opened 2021-08-08
SUBMARINE-965. Users should not have the privilege to create experiments in different namespaces

kevin85421:
### What is this PR for?
In our multi-tenant architecture, users only can create experiments in their namespaces, but the Submarine workbench enables users to specify namespaces.
截圖 2021-08-08 上午10 17 13

This patch aims to 

### What type of PR is it?
[Bug Fix]

### Todos
1. Remove the function `public void setNamespace(String namespace)` in ExperimentMeta.java
2. Update experiment RESTful API (remove the field "namespace" in the JSON file)
3. Add a test case to make sure whether the experiment is in the namespace specified by `ENV_NAMESPACE`.

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-965

**Test1**
* Step1: Use submarine-operator to create a submarine service in the namespace "submarine-user-test"
* Step2: Create an experiment, and check whether the experiment is in the namespace "submarine-user-test" rather than "default".

截圖 2021-08-08 下午11 57 35

**Test2**
* Remove the field "namespace" in the Submarine workbench

截圖 2021-08-08 下午11 57 35

截圖 2021-08-09 上午1 19 46

### Questions:

### DIFF
--- a/submarine-server/server-core/src/test/java/org/apache/submarine/server/rest/ExperimentRestApiTest.java
+++ b/submarine-server/server-core/src/test/java/org/apache/submarine/server/rest/ExperimentRestApiTest.java
@@ -222,6 +222,7 @@ public class ExperimentRestApiTest {
     assertEquals(experimentFinishedTime, experiment.getFinishedTime());
     assertEquals(metaName, experiment.getSpec().getMeta().getName());
     assertEquals(metaFramework, experiment.getSpec().getMeta().getFramework());
+    assertEquals("default", experiment.getSpec().getMeta().getNamespace());
     assertEquals(dockerImage, experiment.getSpec().getEnvironment().getDockerImage());
     assertEquals(kernelChannels, experiment.getSpec().getEnvironment().getKernelSpec().getChannels());
     assertEquals(kernelSpecName, experiment.getSpec().getEnvironment().getKernelSpec().getName());
--- a/submarine-server/server-core/src/test/java/org/apache/submarine/server/rest/ExperimentRestApiTest.java
+++ b/submarine-server/server-core/src/test/java/org/apache/submarine/server/rest/ExperimentRestApiTest.java
@@ -75,7 +75,6 @@ public class ExperimentRestApiTest {
   private static final String experimentStatus = "Succeeded";
   private static final String metaName = "foo";
   private static final String metaFramework = "TensorFlow";
-  private static final String metaNamespace = "fooNamespace";
   private static final String dockerImage = "continuumio/anaconda3";
   private static final String kernelSpecName = "team_default_python_3";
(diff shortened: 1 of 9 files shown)


kevin85421:
@pingsutw Can you help me review this PR? Thanks!

submarine-server/server-core/src/test/java/org/apache/submarine/server/rest/ExperimentRestApiTest.java
```diff
@@ -224,7 +222,6 @@ private void verifyResult(Experiment experiment, String uid) {
     assertEquals(experimentFinishedTime, experiment.getFinishedTime());
     assertEquals(metaName, experiment.getSpec().getMeta().getName());
     assertEquals(metaFramework, experiment.getSpec().getMeta().getFramework());
-    assertEquals(metaNamespace, experiment.getSpec().getMeta().getNamespace());
```
pingsutw, line 227: I think we should assert that the experiment namespace is default. And we should add a new test to make sure the namespace is what we expect if we update `ENV_NAMESPACE`

pingsutw (commented):
other LGTM

kevin85421:
@pingsutw Thank you for your review. I have added the assertion to `ExperimentRestApiTest.java`. In addition, I updated the TODO in this pull request description.

The integration test is very slow in "apache/submarine"; thus, I use the CI result of "kevin85421/hadoop-submarine" to ensure the correctness of this patch.

https://github.com/kevin85421/hadoop-submarine/actions/runs/1113309003

KUAN-HSUN-LI:
@kevin85421 I am confused about running `mvn install -DskipTests` again. I think it will take more time in the integration test.

kevin85421:
@KUAN-HSUN-LI The command `mvn install -DskipTests` is used to upload the latest package to the local repository. To elaborate, the module `submarine-test` has a dependency `submarine-server-core`, and thus we need to upload the latest package. On the other hand, without the command, maven will repost errors.

For more details, you can check these two pull requests.
https://github.com/apache/submarine/pull/695/files
https://github.com/apache/submarine/pull/683/files

Thanks!

KUAN-HSUN-LI:
@kevin85421 Thanks for your reply!
At first, I had mentioned this problem therefore I saved the dependency in the cache. I wished it will use the latest dependency but it didn't work as I expect. For the reason to reduce the CI time and stable tests, I think it is useful to build the submarine in the e2e and k8s jobs. 
(i.e. build -> submarine-e2e  =>  submarine-e2e  and  build -> submarine-k8s  =>  submarine-k8s )

pingsutw (approved):
LGTM

PR #704, opened 2021-08-09
SUBMARINE-966. Correct the job label selector

aeioulisa:
### What is this PR for?
Get empty reply when getting logs because the selector key 'tf-job-name' represent to ExperimentId now, correct it to get the right response.

### What type of PR is it?
[Bug Fix]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-966

image

### Questions:


pingsutw (approved):
+1, thanks.

PR #707, opened 2021-08-13
SUBMARINE-975. Create storageClass for database, minio, tensorboard to control persistentVolume

Kenchu123:
### What is this PR for?

To create persistent volumes, submarine-operator creates persistent volumes with different names depending on the namespace (eg. `submarine-database-pv–default`)

However, we can use storage class to create persistent volumes dynamically and delete them easily.

This will be implemented only on the submarine operator. Besides, only minikube support the current storage class provisioner, and volumes can only be in the host (nfs not supported).

### What type of PR is it?
[Improvement]

### Todos
- [x] Add storage class
- [x] Modify operator to create persistent volume claim with storage class name
- [x] Remove persistent volume creation in operator

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-975

Follow submarine-operator instructions.

https://user-images.githubusercontent.com/17617373/129319157-5923ebcf-cdf4-45ae-b18c-e3cc0eb45b16.mov

### Questions:

### DIFF
--- a/submarine-cloud-v2/helm-charts/submarine-operator/templates/rbac.yaml
+++ b/submarine-cloud-v2/helm-charts/submarine-operator/templates/rbac.yaml
@@ -56,7 +56,6 @@ rules:
       - namespaces
       - jobs
       - serviceaccounts
-      - persistentvolumes
       - persistentvolumeclaims
       - pods/portforward
       - events
@@ -78,8 +77,8 @@ rules:
   - apiGroups:
       - "rbac.authorization.k8s.io"
     resources:
-      - clusterroles
-      - clusterrolebindings
+      - roles
+      - rolebindings
     verbs:
       - "*"
   - apiGroups:
--- a/submarine-cloud-v2/helm-charts/submarine-operator/templates/storageclass.yaml
+++ b/submarine-cloud-v2/helm-charts/submarine-operator/templates/storageclass.yaml
@@ -1,0 +1,45 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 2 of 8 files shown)


kevin85421 (commented):
Thank @Kenchu123 for your contribution! I have a question. Why do we need the file`storageclass.yaml`? In my opinion, the hard-coded StorageClass names cannot support multi-tenancy. To elaborate, in your implementation, two Submarine services in two different namespaces will use the same name "submarine-database-sc" as the name of their DB StorageClass.

KUAN-HSUN-LI:
@Kenchu123 I think you can provide a named template (_helper.tpl file) in the template file. It can create the default value for the provisioner and can be customized by users. With `named template`, users and developers can set their own environment variable like `helm install --set <your env> ...`.
more info: https://helm.sh/docs/chart_template_guide/named_templates/

Kenchu123:
> Thank @Kenchu123 for your contribution! I have a question. Why do we need the file`storageclass.yaml`? In my opinion, the hard-coded StorageClass names cannot support multi-tenancy. To elaborate, in your implementation, two Submarine services in two different namespaces will use the same name "submarine-database-sc" as the name of their DB StorageClass.

Storage class is like an agent for controlling persistent volumes. Take submarine as an example, for submarine servers in different namespaces, they will have different persistent volume claims, so the storage class's provisioner will then create different persistent volumes for each submarine server.
Besides, the storage class is cluster-scoped, so I think components in different namespaces can use the same storage class.

Kenchu123:
> @Kenchu123 I think you can provide a named template (_helper.tpl file) in the template file. It can create the default value for the provisioner and can be customized by users. With `named template`, users and developers can set their own environment variable like `helm install --set <your env> ...`.
> more info: https://helm.sh/docs/chart_template_guide/named_templates/

Thanks. This is what I'm looking for~

kevin85421:
@Kenchu123 Thank you for your explanation! I got the idea. In addition, will you add a named template in this PR?

Kenchu123:
> @Kenchu123 Thank you for your explanation! I got the idea. In addition, will you add a named template in this PR?

I think I'll make a new pull request. The named template needs some tests and document modification.

pingsutw (approved):
+1

PR #714, opened 2021-08-16
SUBMARINE-985. Fix limit on the experiment name

jeff-901:
### What is this PR for?
The hyphens can't be the end of the experiment name too. Update the current regex.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-985

### Questions:


pingsutw (approved):
LGTM

PR #715, opened 2021-08-18
SUBMARINE-982. Install seldon-core

jeff-901:
### What is this PR for?
Install seldon-core and install istio for the seldon-core gateway. Update the document for the installation by script and how to uninstsall.

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-982

### Questions:


jeff-901:
@KUAN-HSUN-LI help me review the code. Thanks.

KUAN-HSUN-LI (approved):
LGTM

lowc1012:
I don't think istio should be used at this time.
Could we only use Traefik if it is feasible?

jeff-901:
> I don't think istio should be used at this time.
> Could we only use Traefik if it is feasible?

Seldon-core currently only supports istio or ambassador for ingress.

pingsutw:
@lowc1012 any suggestion? Seldon-core can't support Traefik.

pingsutw (approved):
+1, cc @lowc1012 Could you help me review it?

lowc1012:
Currently, we use Traefik as the reverse proxy for Submarine.
Istio is a service mesh that provides more powerful features than which Traefik supports.
In my opinion, we need to think more carefully and be familiar with Istio and Envoy If we want to replace Traefik with Istio.

jeff-901:
@KUAN-HSUN-LI  @pingsutw  I move the istio and seldon-core installation to the submarine-serve directory.

PR #717, opened 2021-08-19
SUBMARINE-990. Add helm templates to support storage class configurations

Kenchu123:
### What is this PR for?

From SUBMARINE-975, we can find out that the storage class configurations (provisioner, reclaim policy, etc) is unchangeable.

By adding helm templates (`_helper.tpl`) and adding more values (`values.yaml`), we can use `helm install --set storageClass.provisioner=...` to change the default configurations of the storage class. Therefore, we can easily set up different configurations on different clusters such as kind.

### What type of PR is it?
[Improvement]

### Todos
* [x] - Add `_helpers.tpl` and storageClass fields values
* [x] - Change pvc access mode to `ReadWriteOnce` since local provisioner only support this
* [x] - Modify `README.md`

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-990

For one using minikube, you don't need to make any changes since minikube provisioner is the default.

For one using kind, you need to add `--set storageClass.provisioner=rancher.io/local-path --set storageClass.volumeBindingMode=WaitForFirstConsumer` in `helm install` command.

For one using other storage class provisioners, you can simply set values in `values.yaml` or set with `--set` as above.

ps. This only supports the submarine controlled by the submarine operator.

### Questions:


KUAN-HSUN-LI:
+1

pingsutw (approved):
+1

MortalHappiness:
+1

PR #734, opened 2021-08-31
SUBMARINE-981. Update com.google.guava:guava version

aeioulisa:
### What is this PR for?
Upgrade com.google.guava:guava to version 30.0-jre

A temp directory creation vulnerability exist in Guava versions prior to 30.0 allowing an attacker with access to the machine to potentially access data in a temporary directory created by the Guava com.google.common.io.Files.createTempDir(). The permissions granted to the directory created default to the standard unix-like /tmp ones, leaving the files open. We recommend updating Guava to version 30.0 or later, or update to Java 7 or later, or to explicitly change the permissions after the creation of the directory if neither are possible.

### What type of PR is it?
[Improvement]

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-981

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,7 +110,7 @@
     <jsr305.version>1.3.9</jsr305.version>
     <mockito.version>2.23.4</mockito.version>
     <powermock.version>1.6.4</powermock.version>
-    <guava.version>22.0</guava.version>
+    <guava.version>30.0-jre</guava.version>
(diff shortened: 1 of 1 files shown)


pingsutw (approved):
+1

PR #735, opened 2021-09-01
SUBMARINE-964. Pull notebook image is very time consuming

FatalLin:
### What is this PR for?
Pulling the image of notebook is a quite time consuming process, and it will make user confuse that what the development is waiting for, so we add a new status - Pulling to indicate  the pending is caused by pulling operation.

### What type of PR is it?
Improvement.

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-964
Observe the status in notebook page after creating a notebook.
截圖 2021-09-01 下午6 44 52

### Questions:

### DIFF
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
@@ -31,7 +31,9 @@ import org.apache.submarine.server.environment.EnvironmentManager;
 
 import javax.ws.rs.core.Response;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.submarine.server.notebook.database.NotebookService;
 import org.slf4j.Logger;
@@ -81,7 +83,17 @@ public class NotebookManager {
     String lowerName = spec.getMeta().getName().toLowerCase();
     spec.getMeta().setName(lowerName);
     NotebookId notebookId = generateNotebookId();
+
+    Map<String, String> labels = spec.getMeta().getLabels();
+
+    if (labels == null) {
+      labels = new HashMap<>();
+    }
+    labels.put("notebook-owner-id", spec.getMeta().getOwnerId());
+    labels.put("notebookId", notebookId.toString());
+    spec.getMeta().setLabels(labels);
     Notebook notebook = submitter.createNotebook(spec);
+
     notebook.setNotebookId(notebookId);
     notebook.setSpec(spec);
 
@@ -197,6 +209,13 @@ public class NotebookManager {
     if (spec == null) {
       throw new SubmarineRuntimeException(Response.Status.OK.getStatusCode(),
           "Invalid. Notebook Spec object is null.");
+    }
+    List<Notebook> serviceNotebooks = notebookService.selectAll();
+    for (Notebook notebook: serviceNotebooks) {
+      if (notebook.getSpec().getMeta().getName().equals(spec.getMeta().getName())) {
+        throw new SubmarineRuntimeException(Response.Status.OK.getStatusCode(),
+          "Invalid. Notebook with same name is already existed.");
+      }
     }
   }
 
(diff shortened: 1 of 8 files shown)


pingsutw (approved):
+1, awesome, thanks.

lowc1012 (commented):
@FatalLin  Thanks!
Would it be better to improve below method?
org.apache.submarine.server.submitter.k8s.util.NotebookUtils#processStatus

FatalLin:
@lowc1012 oops I didn't notice the method before, I'll try try the split the codes into NotebookUtils and K8sSubmitter, maybe you could come back and check later.

FatalLin:
> @lowc1012 oops I didn't notice the method before, I'll try try the split the codes into NotebookUtils and K8sSubmitter, maybe you could come back and check later.

@lowc1012 I just checked

FatalLin:
@lowc1012 I just checked the current flow and found out it would be pretty complicate if we move the codes into NotebookUtils.

current flow:
`NotebookManager -> k8sSubmitter -> NotebookUtil`
if we want to move the codes into NotebookUtil, notebookUtil will need to call k8sSubmitter to get further details of pods and event which will cause a calling loop, and I'm don't think it's a good design for us.
I had an alternative solution is moving part of the codes into NotebookManager, like the notebookManager will check is pod waiting due to pulling operation(and this is also a specific case for k8sSubmitter, even yarn is no longer supported).
How do you think about these?

lowc1012:
@FatalLin
As we discussed offline, we can raise another issue to refactor K8sSubmitter and NotebookManager.
Thanks for your suggestion!

pingsutw (approved):

FatalLin:
@kevin85421 @pingsutw @lowc1012  Hi guys, thanks for the discussion in the last two days!
I just pushed the last commit of this PR, and this should be the final version of it.
per discussion with Kai-Hsun, we added a new Map property into Notebook Meta Class, which is also could be leveraged later if we allow user to enter their own label from front-end.
We put the notebookId into the label properties, and all the operation depends on the name has been modified, they are depending on the notebookId now.
I also added the check while notebook creating operation to make sure the notebook name is unique.

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/K8sSubmitter.java
```diff
@@ -429,10 +429,30 @@ public Notebook findNotebook(NotebookSpec spec) throws SubmarineRuntimeException
...
+        String podLabelSelector = String.format("%s=%s", NotebookCR.NOTEBOOK_ID,
+            spec.getMeta().getLabels().get(NotebookCR.NOTEBOOK_ID).toString());
+
+        V1PodList podList = coreApi.listNamespacedPod(namespace, null, null, null, podLabelSelector,
+            null, null, null, null);
+        String podName = podList.getItems().get(0).getMetadata().getName();
+        String fieldSelector = String.format("involvedObject.name=%s", podName);
+        V1EventList events = coreApi.listNamespacedEvent(namespace, null, null, fieldSelector,
+            null, null, null, null, null);
+        V1Event lastestEvent = events.getItems().get(events.getItems().size() - 1);
```
lowc1012, line 449: latestEvent?

submarine-server/server-core/src/main/java/org/apache/submarine/server/notebook/NotebookManager.java
```diff
@@ -81,7 +83,17 @@ public Notebook createNotebook(NotebookSpec spec) throws SubmarineRuntimeExcepti
...
     spec.getMeta().setName(lowerName);
     NotebookId notebookId = generateNotebookId();
+
+    Map<String, String> labels = spec.getMeta().getLabels();
+
+    if (labels == null) {
+      labels = new HashMap<>();
+    }
+    labels.put("notebook-owner-id", spec.getMeta().getOwnerId());
+    labels.put("notebookId", notebookId.toString());
```
lowc1012, line 93: Use "notebook-id" ?
ref: [syntax-and-character-set](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

FatalLin:
@lowc1012 good call! both suggestion had been integrated into this PR.

pingsutw (approved):
Thank you @FatalLin @lowc1012.

lowc1012 (approved):
+1, Thanks @FatalLin

PR #739, opened 2021-09-07
SUBMARINE-1009. Support mypy for Submarine SDK

KUAN-HSUN-LI:
### What is this PR for?
[mypy](https://github.com/python/mypy) is an optional static type checker for Python.
With `mypy` we can find bugs in our programs without even running them.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1009

ci pass

### Questions:


pingsutw (approved):
LGTM, thanks @KUAN-HSUN-LI

PR #744, opened 2021-09-12
SUBMARINE-1012. Submarine CR status is inconsistent with actual state

MortalHappiness:
### What is this PR for?
When creating the submarine using the operator, the status of the submarine CR shows mlflow failed due to time-out processing. However, the mlflow deployment is running.

#### Cause:
Submarine operator treat progressing error, ex. ProgressDeadlineExceeded, as a fatal error. However, the deployment controller will continue to process failed deployment, so the deployment may be running in several minutes. This causes inconsistent CR status with the actual state.

#### Solution:
Does not treat progressing error as a fatal error

#### References:
https://pkg.go.dev/k8s.io/api/apps/v1#DeploymentSpec

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1012

### Questions:


MortalHappiness:
@pingsutw @Kenchu123 Can you help me review this PR? Thanks.

Kenchu123:
LGTM.

pingsutw (approved):
+1

PR #750, opened 2021-09-15
SUBMARINE-1025. Use minio client in init container to create bucket

jeff-901:
### What is this PR for?
Migrate the minio logic to init container for submarine and mlflow pod. If the init container fails, let it restart.

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1025

Original e2e test.

### Questions:


jeff-901:
@KUAN-HSUN-LI @pingsutw Please helm me review the code, thanks.

KUAN-HSUN-LI:
The init container will always restart until minio pod is running. If minio pod fails, both submarine and mlflow pods will still restart the init container. Is this a serious problem? @pingsutw

pingsutw (approved):
LGTM, Thanks.

PR #753, opened 2021-09-16
SUBMARINE-1028. Extend timeout for integration tests

KUAN-HSUN-LI:
### What is this PR for?
Sometimes GitHub Actions' server is not stable, then it will take around 20 minutes and lead to timeout of integration tests.

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1028

### Questions:

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -28,7 +28,7 @@ env:
 jobs:
   submarine-e2e:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     services:
       mysql:
         image: mysql:5.7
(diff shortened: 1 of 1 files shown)


pingsutw (approved):
+1, Thanks for fixing it.

cxorm:
Overall LGTM, just one question from my side.

Could you please share the reason that setting the new timeout-minutes to 60 instead of other time ?

KUAN-HSUN-LI:
@cxorm If there isn't a restart happen, test-k8s needs around 14 minutes, and test-e2e needs around 9 minutes. Usually restart will take 20~30 minutes therefore I think 60 minutes is a proper setting. Additionally, the integration test in python workflow is also using 60 minutes for the timeout.

cxorm (approved):
Nice catch, +1

PR #796, opened 2021-11-11
SUBMARINE-1078. Retire Spark security module from submarine project

yaooqinn:
### What is this PR for?

Retire Spark security module from the submarine project

### What type of PR is it?
 
Refactoring

### Todos

None

### What is the Jira issue?

SUBMARINE-1078

Passing GA

### Questions:


xunliu (approved):
Spark security module originally belonged to another project, kyuubi.
 now kyuubi It has become a new incubator project https://github.com/apache/incubator-kyuubi, so the Spark security module needs to be migrated from submarine to kyuubi.

PR #809, opened 2021-11-23
SUBMARINE-1065. Fix Manager-User Cannot Submit Changes

atosystem:
### What is this PR for?
1. Fix the bug that workbench manager-user cannot submit changes
The error is that the password validator still working in edit mode, but the password value in edit mode is a hash value. Hence, the validation won't pass.
2. Change the order of `Manager` and `Model` on the side menu of Workbench.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
[https://issues.apache.org/jira/browse/SUBMARINE-1065](https://issues.apache.org/jira/browse/SUBMARINE-1065)

bugfixed
> This animation is recorded before the update of sidebar order.

### Questions:


pingsutw (approved):
+1, thanks.

PR #812, opened 2021-11-27
SUBMARINE-1092. Use Model Management REST API in Model Management Page

andybi7676:
### What is this PR for?
remove the static part in the model main page and use REST API to fetch data from the database. 

### What type of PR is it?
Improvement
### Todos
* [x] - model main page fetching
* [x] - loading icon when fetching 

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-1092

https://user-images.githubusercontent.com/59821107/143672967-1bd64dfc-24fb-4363-942b-44bf651bb4db.mp4

### Questions:

### DIFF
--- a/submarine-workbench/workbench-web/src/app/pages/workbench/model/model-home/model-home.component.ts
+++ b/submarine-workbench/workbench-web/src/app/pages/workbench/model/model-home/model-home.component.ts
@@ -47,12 +47,11 @@ export class ModelHomeComponent implements OnInit {
     this.modelService.fetchModelList().subscribe((res) => {
       this.modelCards = res;
       this.onDisplayModelCards = this.modelCards.map(card => card);
-      let tags = [];
+      let tags_set = new Set();
       this.modelCards.map((card) => {
-        Array.prototype.push.apply(tags, card.tags);
+        card.tags.forEach(tags_set.add, tags_set);
       });
-      let tags_set = new Set(tags);
-      tags = Array.from(tags_set);
+      let tags = Array.from(tags_set);
       this.listOfTagsOption = tags.map((tag) => ({ "label": String(tag), "value": String(tag)}));
       this.isModelCardsLoading = false;
     });
--- a/submarine-workbench/workbench-web/src/app/pages/workbench/model/model-home/model-home.component.ts
+++ b/submarine-workbench/workbench-web/src/app/pages/workbench/model/model-home/model-home.component.ts
@@ -17,9 +17,9 @@
  * under the License.
  */
 
-import { Component, OnInit } from '@angular/core';
+import { Component, ComponentFactoryResolver, OnInit } from '@angular/core';
 import { ModelInfo } from '@submarine/interfaces/model-info';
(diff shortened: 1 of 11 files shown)


andybi7676:
@KUAN-HSUN-LI Could you please help me review this PR?

submarine-workbench/workbench-web/src/app/pages/workbench/model/model-home/model-home.component.ts
```diff
@@ -38,80 +39,23 @@ export class ModelHomeComponent implements OnInit {
...
+  fetchModelCards = () => {
+    this.modelService.fetchModelList().subscribe((res) => {
+      this.modelCards = res;
+      this.onDisplayModelCards = this.modelCards.map(card => card);
+      let tags = [];
+      this.modelCards.map((card) => {
+        Array.prototype.push.apply(tags, card.tags);
+      });
+      let tags_set = new Set(tags);
+      tags = Array.from(tags_set);
```
KUAN-HSUN-LI, line 55: We don't need to check the uniqueness of the tags again. DB has made it a primary key.

KUAN-HSUN-LI (changes requested):
Others, LGTM. Thanks

submarine-workbench/workbench-web/src/app/pages/workbench/model/model-home/model-home.component.ts
andybi7676: But different models may have the same tag, right?

KUAN-HSUN-LI: I mistook of thinking this is inside the card element.But just add the element into set and then dump the set to the array will be better? (Skip one step)

andybi7676: Oh, yes. This might be a better solution. I'll resolve it, thanks!

andybi7676: I took a look at JavaScript Set. It turns out that JS Set does not have built-in functionality for union operation. So I think the previous implementation might be the simplest as well. 
image
This figure shows that if we want to union two sets, we still need to convert the sets to an array and then regenerate a new set. 
Another solution is to add the elements one by one. However, this is not more efficient and requires more code.

KUAN-HSUN-LI: `let tags_set = new Set();`
`this.modelCards.map((card) => {
        tags_set.add(card.tags);
});`
`tags = Array.from(tags_set);`
This is what I mean. Is this feasible?

andybi7676: `Set.prototype.add` will treat the input as a value, so the result of the above code would be a set of arrays in it. 

An example:
image

KUAN-HSUN-LI (approved):
LGTM, Thanks

PR #822, opened 2021-12-06
SUBMARINE-1105. implementation of the base frame of k8s-agent

FatalLin:
### What is this PR for?
just like we discussed earlier, we would like to turn submarine into a more k8s specialize solution; thus I build the base frame k8s-agent in this PR, which is responsible for tracking different types of custom resource we created - tfjob, pytorch job, and note book. 
### What type of PR is it?
Feature

### Todos
* implement the customResourceHandler for each custom resource.
* add the building process into current building progress.

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1105
For now the codes is quite simple, I think we could add the unit test once the customResource handler is done.
Screen Shot 2021-11-30 at 11 06 09 PM
Screen Shot 2021-11-30 at 11 06 24 PM

### Questions:
* Does this need new documentation? Yes (Design Doc)


bin/submarine-daemon.sh
```diff
@@ -120,9 +120,9 @@ function start() {
 
   initialize_default_directories
 
-  echo "SUBMARINE_SERVER_CLASSPATH: ${SUBMARINE_SERVER_CLASSPATH}" >> "${SUBMARINE_SERVER_LOGFILE}"
+  echo "SUBMARINE_SERVER_CLASSPATH: ${SUBMARINE_APP_CLASSPATH}" >> "${SUBMARINE_SERVER_LOGFILE}"
```
pingsutw, line 123:
```suggestion
  echo "SUBMARINE_APP_CLASSPATH: ${SUBMARINE_APP_CLASSPATH}" >> "${SUBMARINE_SERVER_LOGFILE}"
```

submarine-server/server-submitter/submarine-k8s-agent/src/main/java/org/apache/submarine/server/k8s/agent/handler/CustomResourceHandler.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getCrName() {
+        return crName;
+    }
+
+    public void setCrName(String crName) {
+        this.crName = crName;
+    }
+    
+    
+
```
pingsutw, line 73:
```suggestion
    
    
```suggestion
```
```

pingsutw (commented):
Others LGTM, thanks

pingsutw (approved):
+1

PR #823, opened 2021-12-06
SUBMARINE-1100. (Experiment) Add Istio sidecar but keep Traefik router

woodcutter-eric:
### What is this PR for?

Keep Traefik as the edge router but use sidecar features in Istio. It it just an experiment.

### What type of PR is it?
[Feature]

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1100?filter=myopenissues

### Questions:


woodcutter-eric:
I still have to verify how to test. Please do not merge temporarily!

woodcutter-eric:
Unfortunately, you can only deploy all pods in the same namespace now (default), or the test will fail.

PR #834, opened 2021-12-16
SUBMARINE-1126. Let Jupyter Notebook save user setting after pod restarted.

cdmikechen:
### What is this PR for?
Jupyter Notebook have created a workspace volume `/home/jovyan/workspace` , but the user settings are saved in `/home/jovyan/.jupyter/lab/user-settings/@jupyterlab/`. 
So that after pod restart, submarine will lost user settings. I think we should add a new volume to save user settings in `/home/jovyan/.jupyter/`.

### What type of PR is it?
Improvement

### Todos
* [x] - Add a new PVC named `notebook-pvc-user-*` when creating notebook pod CRD

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1126

There are no new test cases right now.

image

### Questions:
* Does this need new documentation? Maybe No


KUAN-HSUN-LI (approved):
LGTM. @jeff-901 Can you also take a look? Maybe we can also save the experimentId into the PVC?

pingsutw (approved):
+1

PR #835, opened 2021-12-16
SUBMARINE-1125. Fix Conda Version match error

cdmikechen:
### What is this PR for?
The Conda version in jupyter-notebook-0.7.0-SNAPSHOT docker image is 4.11.0 now. But in Submarine it still check version from 4.0.1 to 4.10.10, so that when notebook pod start, it report error like this:
image

### What type of PR is it?
Bug Fix 

### Todos
* [x] - Replace max version to 4.11

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1125

Can test in new docker image.

### Questions:


KUAN-HSUN-LI (approved):
+1

PR #839, opened 2021-12-19
SUBMARINE-1099. Replace Traefik with Istio in helm-charts

chiaoymir:
### What is this PR for?

Replace Traefik with Istio as gateway. 

### What type of PR is it?
[Feature]

### Todos
* [x] - Fix static files
* [ ] - Fix minio request
* [ ] - Fix experiment request 

### What is the Jira issue?

https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1099?filter=allissues

istio-ingressgateway:
Screen Shot 2021-12-19 at 7 01 13 PM

submarine-minio-service:
Screen Shot 2021-12-19 at 7 01 53 PM

dashboard:
Screen Shot 2021-12-20 at 6 56 55 AM

notebook:
Screen Shot 2021-12-20 at 6 57 11 AM

### Questions:


chiaoymir:
Unfortunately, I failed to completely replace Traefik with Istio, some endpoints seem broken. However, I don't understand why. Therefore, I think I need to hand off this issue to another person.

cdmikechen:
Why not user `helm dependency` to import `istio`?
Maybe we should keep the possibility of two kinds of ingress coexisting, and let users choose their own suitable ingress. In addition, if the user's k8s platform already has its own ingress controller, there is no relevant solution at present.

PR #845, opened 2021-12-21
SUBMARINE-1074. S3 client for server

jeff-901:
### What is this PR for?
Add s3 minio client in server.

### What type of PR is it?
Feature

### Todos
* [x] - list artifact paths by experiment id
* [x] - upload an artifact
* [x] - download an artifact
* [x] - delete artifacts under a experiment id

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1074

### Questions:

### DIFF
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
- 
+
 package org.apache.submarine.server.s3;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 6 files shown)


submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
```diff
@@ -0,0 +1,175 @@
...
+   * Download an artifact
+   *
+   * @param path artifact path
+   * @return an array of byte
+   */
+  public byte[] downloadArtifact(String path) {
+    InputStream is;
+    byte[] buffer;
+    int b;
+    try {
```
pingsutw, line 109: It's better to use the [try-with-resources Statement
](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

jeff-901: ok

pingsutw (approved):
@jeff-901 Thanks, LGTM.
btw, we should define an interface or manager to write to different file systems(s3, GCP).
We can do it in separate tickets

PR #846, opened 2021-12-21
SUBMARINE-1154. Add pytorchjobs/finalizers and tfjobs/finalizers resources RBAC

cdmikechen:
### What is this PR for?
Add pytorchjobs/finalizers and tfjobs/finalizers resources to fix the problem that the experiment pod cannot be created normally

### What type of PR is it?
Bug Fix 

### Todos
* [x] - Add pytorchjobs/finalizers and tfjobs/finalizers resources RBAC

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1154

After the CRD is created, we can check whether the pod starts normally.

### Questions:


KUAN-HSUN-LI (approved):
+1

PR #847, opened 2021-12-21
SUBMARINE-1135. Implement CLI Config Command

atosystem:
### What is this PR for?

Implement `submarine config` command to set *_port_* and *_hostname_* for API connection.
Usage:
`submarine config init` : init config file (restore to default)
`submarine config list` : list config content
`submarine config set <param> <value>` : set param = value in config
`submarine config get <param>` : get param

Examples:
1. Change API port to `8080`
`submarine config set connection.port 8080`
2. Change hostname to `127.0.0.1`
`submarine config set connection.hostname 127.0.0.1`

### What type of PR is it?
[Feature]

### Todos
None

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1135

python unit test is implemented

### Questions:

### DIFF
--- a/submarine-sdk/pysubmarine/setup.py
+++ b/submarine-sdk/pysubmarine/setup.py
@@ -44,7 +44,7 @@ setup(
         "click==8.0.3",
         "rich==10.15.2",
         "dacite==1.6.0",
-        "dataclasses==0.8",
+        "dataclasses>=0.6",
         "pyaml==21.10.1",
     ],
     extras_require={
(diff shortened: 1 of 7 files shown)


atosystem:
@KUAN-HSUN-LI Thanks!

KUAN-HSUN-LI (commented):
Currently, the Java server is using another config. Maybe we should migrate these two configs; otherwise, it will be confusing for users to use this command.

PR #848, opened 2021-12-23
SUBMARINE-1139. add internal api for status sync

FatalLin:
### What is this PR for?
I added an internal service in submarine-server which allow the agent which is appended on our custom resource object to sync their tracking result to database, also push the status change via web-socket.

### What type of PR is it?
Feature

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1139
the unit test should be added later.
flow chart

### Questions:


submarine-server/server-api/src/main/java/org/apache/submarine/server/api/Submitter.java
```diff
@@ -44,10 +44,11 @@
   /**
    * Create experiment with spec
    * @param spec experiment spec
+   * @param id experimentId 
    * @return object
    * @throws SubmarineRuntimeException running error
    */
-  Experiment createExperiment(ExperimentSpec spec) throws SubmarineRuntimeException;
+  Experiment createExperiment(ExperimentSpec spec, String experimentId) throws SubmarineRuntimeException;
```
pingsutw, line 51: We already have `experimentId` in `ExperimentSpec`. 
Check `ExperimentSpec` -> `ExperimentMeta` -> `experimentId`

FatalLin: oh i haven't notice before, will recover it.

FatalLin: the code has been rollback.

pingsutw (approved):

PR #854, opened 2021-12-31
SUBMARINE-1170 Add traefik role verbs to create/update/delete CRD

cdmikechen:
### What is this PR for?
We can add traefik role verbs to create/update/delete CRD, so that new user/RoleBinding ServiceAccount can create or delete IngressRoute.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Add create/update/delete verbs in traefik role

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1170

We can bind other users / roles to test whether we have permission to create CRDs

No

### Questions:


pingsutw (approved):
+1, thanks

PR #857, opened 2022-01-02
SUBMARINE-1163. Remove mlflow

jeff-901:
### What is this PR for?
Remove mlflow from submarine

### What type of PR is it?
Feature

### Todos
* [x] - Remove mllfow from sdk 
* [x] - Update document

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1163

### Questions:


cdmikechen:
There are some codes about mlflow in `K8sSubmitter`, should we need to remove these?

jeff-901:
> There are some codes about mlflow in `K8sSubmitter`, should we need to remove these?

We decide to only remove mlflow in python sdk, so the part of mlflow in `K8sSubmitter` will remain.

pingsutw (approved):

PR #866, opened 2022-01-12
SUBMARINE-1180. Fix the hibernate error in server

KUAN-HSUN-LI:
### What is this PR for?
Fix the bugs in the server.
1. hibernate dependency error
2. split interruptedexception from Exception

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1180

### Questions:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
     <minio.version>7.1.4</minio.version>
 
     <cglib.version>3.3.0</cglib.version>
-    <hibernate.version>5.6.0.Final</hibernate.version>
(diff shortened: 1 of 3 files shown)


cdmikechen:
I have a question: if we use Mybatis, why do we import Hibernate?

KUAN-HSUN-LI:
We have discussed the database issue about half a year ago but I think it is suitable to use mybatis now. I will discuss whether to remove it in the next meeting.

pingsutw (approved):

PR #875, opened 2022-01-24
SUBMARINE-1184. Get namespace via Downward API

MortalHappiness:
### What is this PR for?
Get the current namespace with the downward API.

In the beginning, I tried to get the current namespace by reading the file `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. However, not all Kubernetes environments have this file. Therefore I decided to use the downward API to get the current namespace.

References:
+ https://stackoverflow.com/questions/31557932/how-to-get-the-namespace-from-inside-a-pod-in-openshift
+ https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/user-guide/downward-api.md

### What type of PR is it?
[Improvement]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1184

### Questions:


pingsutw (approved):
+1, thanks

PR #878, opened 2022-02-14
SUBMARINE-1187.  Refine the interface of internal service and keep more details into db

FatalLin:
### What is this PR for?
per discussion with @pingsutw, we would like to persist more detail data into database which allow us to provide to users when they check the detail, therefore I refine the logic when handler receiving the events, it will align the same logic in k8ssubmitter to rebuild the corresponding object, and pass to internal service, then the internal service persist it into database.

### What type of PR is it?
Improvement
### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1187
existed unit tests.
same of: https://github.com/apache/submarine/pull/877
### Questions:


submarine-server/server-core/pom.xml
```diff
@@ -437,6 +437,11 @@
           <artifactId>jackson-annotations</artifactId>
         </exclusion>
       </exclusions>
+    </dependency>
+	<dependency>
```
jiwq, line 441: indentation nit~

submarine-server/server-core/pom.xml
```diff
@@ -437,6 +437,11 @@
...
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+      <version>2.10.8</version>
```
jiwq, line 444: - should better move the dependency to center management
- the artifact's version should define in property

submarine-server/server-core/src/main/java/org/apache/submarine/server/experiment/database/entity/ExperimentEntity.java
```diff
@@ -19,6 +19,7 @@
 
 package org.apache.submarine.server.experiment.database.entity;
 
+
```
jiwq, line 22: redundant blank line

pingsutw (approved):
+1, Thank you @FatalLin and @jiwq

PR #886, opened 2022-02-22
SUBMARINE-1199. add style check in submarine-agent module

FatalLin:
### What is this PR for?
add code style check plugin in agent module and polish the code in the same time.

### What type of PR is it?
Improvement

### Todos
NA

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1199
this change should not effect the code.
### Questions:


pingsutw (approved):
+1, thanks

PR #894, opened 2022-03-02
SUBMARINE-1099. Replace Traefik with Istio

MortalHappiness:
### What is this PR for?
Replace Traefik with Istio.

### What type of PR is it?
[Improvement]

### Todos
* [x] - Ensure mlflow and tensorboard can be accessed
* [x] - Ensure notebook can be accessed

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1099

### Questions:


MortalHappiness:
@pingsutw @Kenchu123 This PR can run successfully in my local machine after I set up a kind cluster and build the docker images and execute `start-submarine.sh` manually. However, in GitHub Actions all pods remain pending after a long period of time so the tests failed. I cannot find the reason. Can you help me?

Screenshot from 2022-03-02 18-02-16

Kenchu123:
I have tried to run the code on my computer but it still works well.

noidname01:
I think I found the problem.
2022-03-04 17-07-57 的螢幕擷圖
Those pods are pending because of no more CPU resources can use. I think provide a larger node size should fix this problem.

noidname01:
Here is the total usage of CPU.

2022-03-04 18-19-03 的螢幕擷圖
2022-03-04 18-19-19 的螢幕擷圖

PR #897, opened 2022-03-10
SUBMARINE-1212. fine tune crud operation flow of notebook to align experiment flow

FatalLin:
### What is this PR for?
After we completed the refactoring of experiment operation flow, we would like to align the same pattern on notebook flow; In that case, both the experiment and notebook flow has the same flow. Besides, we also could collect all the status change logic into agent.

### What type of PR is it?
Improvement

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1212
should passed all existed tests.
### Questions:


pingsutw (approved):
lgtm, thanks

KUAN-HSUN-LI (approved):
Thanks

PR #898, opened 2022-03-12
SUBMARINE-1215. Fix java.lang.NullPointerException in K8sSubmitter.getInfo() method

cdmikechen:
### What is this PR for?
When there is no mlflow, a `NullPointerException` will appear in the background. This PR is mainly to fix the background problem.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Use Optional to avoid NullPointerException

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1215

After fixing the background problem, the button will always turn when there is no mlflow in the k8s. I don't know if the front also needs to deal with it.
image

No

### Questions:


KUAN-HSUN-LI (approved):
LGTM, Thanks.

pingsutw (approved):

PR #910, opened 2022-03-24
add more event type to handle to present notebook status correctly

FatalLin:
### What is this PR for?
We found out the root cause of incorrect status display is there's couple event type we skip before in notebook handler's processing, so I added them in the switch case in this PR. For now on, the notebook status would be
`creating -> pulling -> running -> failed -> pulling -> running -> failed -> pulling .....`.
I also noticed that running status of second or third time retry may missed, but looks like this is some kind of bug from k8s, because running state is also be jump through when I used `kubectl describe` to get the status of pods.
btw, I can't simulate the deleting failed case, so the issue hadn't been handled in this pr.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1022
should pass all existing tests.

https://user-images.githubusercontent.com/5687317/159927060-0b5e9587-a399-450a-ba1f-4dc6bcfaa765.mov

### Questions:
* Does this need new documentation?No


pingsutw (approved):
Thanks

PR #911, opened 2022-03-26
SUBMARINE-1227. Fix the operator nil error by the conflict of deployment name

cdmikechen:
### What is this PR for?
When there has been a deployment called `submarine-server` in the same namespace,  create a new submarine CR in this namespace will cause nil exception. After several crash downs of the operator, the operator will not be available.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Handle the error that the object is nil when get deployment by name

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1227

No

### Questions:


MortalHappiness (approved):
LGTM

PR #919, opened 2022-04-04
SUBMARINE-1251. fix patch type and add error handling

KUAN-HSUN-LI:
### What is this PR for?
In the previous PR #915, the patch method is not fixed since the `InvalidSpecException` is sending status code 200.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1251

### Questions:


pingsutw (approved):
LGTM

PR #922, opened 2022-04-06
SUBMARINE-1204. Notebook names cannot start with "submarine"

FYRichie:
### What is this PR for?
The name of the notebook cannot start with "submarine". Otherwise, the service name will conflict with the other services that the operator created.
For example, if the name of the notebook is "submarine-server", then it will conflict with the service "submarine-server" that the operator created.
This PR aims to avoid this problem by adding the notebookId in front of notebookName at k8sSubmitter, but the notebookId will not display on frontend.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1204?filter=allopenissues
Run the submarine service, and check `kubectl get notebook -n submarine-user-test`, if the notebook name in k8s is an <id>-<name> format, and the frontend only displays notebook name, then it is correct.
### Questions:


KUAN-HSUN-LI:
@FYRichie Please fix the error. Thanks

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/K8sSubmitter.java
```diff
@@ -466,7 +466,9 @@ private boolean isDeploymentAvailable(String name) throws ApiException{
   @Override
   public Notebook createNotebook(NotebookSpec spec, String notebookId) throws SubmarineRuntimeException {
     Notebook notebook;
-    final String name = spec.getMeta().getName();
+    // final String name = spec.getMeta().getName();
```
pingsutw, line 469: nit: should we remove this

pingsutw (approved):

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/K8sSubmitter.java
FYRichie: @pingsutw I will remove this

FYRichie:
@MortalHappiness @FYRichie Can you help me review my code?

MortalHappiness (approved):
LGTM

PR #925, opened 2022-04-08
SUBMARINE-1254 fix cannot subscriptable error

KUAN-HSUN-LI:
### What is this PR for?
In Python3, map, filter... functions return an object, not a list. We need to convert the object to the list.

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1254

### Questions:


pingsutw (approved):
Thanks

PR #939, opened 2022-04-19
SUBMARINE-1255. Add XGBoost support to MLJob

JackLiu00521:
### What is this PR for?
Add XGBoost support to MLJob

### What type of PR is it?
Feature

### Todos
* [x]  update ExperimentSpecParser
* [x]  create XGBoost Folder in Model Folder
* [ ]  add XGBoost button in Workbench

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1255

### Questions:


cdmikechen:
@pingsutw 
I'm thinking whether we can abstract the relevant services into interfaces, so that we can integrate faster in the future.

PR #958, opened 2022-05-17
SUBMARINE-1270. add server-database's test into git hub action and extract the coverage analyze  source from all jar with submarine prefix

FatalLin:
### What is this PR for?
The coverage is still not worked since last PR, so I dig the coverage script and try to find more information.
I found out the original script only extract the class form submarine-all* jar, but a huge numbers of module's jar is put under lib directory, so all of them won't be scanned.
In the mean time, I also noticed the github action didn't execute the unit tests of  submaine-database, so I add them into flow as well.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-1270

should passed existed tests.
image

### Questions:
* Does this need new documentation?No


pingsutw (approved):
lgtm, thanks

PR #966, opened 2022-06-12
SUBMARINE-1282. add the codes of experiment prehandler module

FatalLin:
### What type of PR is it?
[Bug Fix | Improvement | Feature | Documentation | Hot Fix | Refactoring]

### Todos
* [ ] - Task

### Questions:
* Do the license files need updating? Yes/No
* Are there breaking changes for older versions? Yes/No
* Does this need new documentation? Yes/No

### DIFF
--- a/submarine-experiment-prehandler/prehandler_main.py
+++ b/submarine-experiment-prehandler/prehandler_main.py
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+from fs_prehandler.hdfs_prehandler import HDFSPreHandler 
+
+if __name__ == "__main__":
+    fsPrehandler = HDFSPreHandler()
+    fsPrehandler.process()
+
+    
--- a/bin/experiment-prehandler.sh
+++ b/bin/experiment-prehandler.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


submarine-experiment-prehandler/prehandler_main.py
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+from fs_prehandler.hdfs_prehandler import HDFSPreHandler 
+
+if __name__ == "__main__":
+    fsPrehandler = HDFSPreHandler()
```
pingsutw, line 19: We could add a new function `get_prehandler` to create corresponding filesystems. Create an HDFS handler if the uri prefix is `hdfs://`or an s3 handler if the uri prefix is`s3://`. What do you think?
```suggestion
    fsPrehandler = get_prehandler(uri=uri)
```

FatalLin: I think its a good approach if we leverage some kind of factory pattern here, but I thought maybe we could use another environment to determine the instance we get here? Because user may need setup different configurations for different file systems, and our UI also need to provide specific  fields for them.

cdmikechen:
```dockerfile
RUN python3 -m pip install --upgrade pip \
    && pip3 install fsspec pyarrow
```
Maybe add `no-cache-dir` to reduce image size? Otherwise, there will be cache files downloaded by pip in the image which we don't need.

cdmikechen:
```python
class HDFSPreHandler(FsPreHandler):
    def __init__(self):
        self.hdfs_host=os.environ["HDFS_HOST"]
        self.hdfs_port=int(os.environ["HDFS_PORT"])
        self.hdfs_source=os.environ["HDFS_SOURCE"]
        self.dest_path=os.environ["DEST_PATH"]
        self.enable_kerberos=os.environ["ENABLE_KERBEROS"]
```
Do we need to consider Hadoop HA? In HA case, the host may be mapped to `dfs.nameservices`, so that we may need to upload `core-sit.xml` and `hdfs-site.xml` files (of course, this is a Java way, will Python simplify it?).
And sometimes some Hadoop host names may be set independently by the internal server in `/etc/hosts` rather than obtained in DNS. Should we add `hostaliases` to the pod/deployment resource of submarine?

FatalLin:
> 

cool, will do.

FatalLin:
> ```dockerfile
> RUN python3 -m pip install --upgrade pip \
>     && pip3 install fsspec pyarrow
> ```
> 
> Maybe add `no-cache-dir` to reduce image size? Otherwise, there will be cache files downloaded by pip in the image which we don't need.

cool, will do.

FatalLin:
> ```python
> class HDFSPreHandler(FsPreHandler):
>     def __init__(self):
>         self.hdfs_host=os.environ["HDFS_HOST"]
>         self.hdfs_port=int(os.environ["HDFS_PORT"])
>         self.hdfs_source=os.environ["HDFS_SOURCE"]
>         self.dest_path=os.environ["DEST_PATH"]
>         self.enable_kerberos=os.environ["ENABLE_KERBEROS"]
> ```
> 
> Do we need to consider Hadoop HA? In HA case, the host may be mapped to `dfs.nameservices`, so that we may need to upload `core-sit.xml` and `hdfs-site.xml` files (of course, this is a Java way, will Python simplify it?). And sometimes some Hadoop host names may be set independently by the internal server in `/etc/hosts` rather than obtained in DNS. Should we add `hostaliases` to the pod/deployment resource of submarine?

well, as a client, I don't think Submarine need to worry about the server side behavior here. If the cluster has any HA mechanism and it also knows there's any client would connect the cluster service,  the cluster administrator should find out how to provide stable serving when the swtich-over happened.

PR #969, opened 2022-06-19
SUBMARINE-1261. Put configurations into Kubernetes Configmap

Ray02250418:
### What is this PR for?
Collect configurations in conf/submarine-sit.xml, common-utils/SubmarineConfiguration.java, and common-utils/SubmarineConfVars into Kubernetes Configmap.

### What type of PR is it?
Improvement

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1261?filter=allissues

### Questions:


Ray02250418:
@KUAN-HSUN-LI Please help me review the code

cdmikechen:
```java
    // Debug
    LOG.info("////////////////////    setupJettyServer Debug    //////////////////");
    LOG.info("SUBMARINE_SERVER_JETTY_THREAD_POOL_MAX: " +
        Integer.toString(conf.getInt(SubmarineConfVars.ConfVars.SUBMARINE_SERVER_JETTY_THREAD_POOL_MAX)));
```
If we explicitly use debug log in this part, I suggest changing the log level to debug, for example:
```java
if (LOG.isDebugEnabled()) {
    LOG.debug("xxxxx");
}
```

cdmikechen:
```yaml
jdbc.url: "jdbc:mysql://127.0.0.1:3306/submarine?useUnicode=true&characterEncoding=UTF-8&autoReconnect=true&allowMultiQueries=true&failOverReadOnly=false&zeroDateTimeBehavior=convertToNull&useSSL=false&serverTimezone=UTC&useTimezone=true&useLegacyDatetimeCode=true"
```
If we choose to replace XML configuration based on ConfigMap, I suggest changing `127.0.0.1` to `MySQL` service name. And it is better to provide a method/solution so that developers can easily develop and debug on their computers.

cdmikechen:
```java
ConfVars(String varName, VarType type) {
  switch(type) {
    case STRING:
      this.varName = varName;
      this.varClass = String.class;
      if (varName == "submarine.server.ssl.key.manager.password" ||
          varName == "submarine.server.ssl.truststore.path" || 
          varName == "submarine.server.ssl.truststore.type" ||
          varName == "submarine.server.ssl.truststore.password"
      ) {
        this.stringValue = null;
      } else {
        this.stringValue = System.getenv(varName);
      }
      this.intValue = -1;
      this.floatValue = -1;
      this.longValue = -1;
      this.booleanValue = false;
      break;
```
I don't think it is a good way to remove the default value from the method parameters and use `if/else` to judge. 
Maybe it is a better choice for us to convert xml and variables by `enum` or `properties` files?

PR #970, opened 2022-06-21
SUBMARINE-1286. Install grafana

jeff-901:
### What is this PR for?
Install grafana by using grafana original image

### What type of PR is it?
Feature

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1286

go to http://{istio ingress ip}/grafana

image

### Questions:


cdmikechen:
Can we set `grafana.ini` as an entry in the CRD to let user define their own grafana configurations.
At the same time, I think the storage of grafana can also be migrated to MySQL, so that we can unify the storages, otherwise the storage of grafana is separated from other services of submarine.

PR #985, opened 2022-08-09
SUBMARINE-976. Configure ingress routing for services across different namespaces

joshvictor1024:
### What is this PR for?

Configure Istio virtual service routing for submarine services across different namespaces.

For example, a submarine workbench created by submitting `Submarine` custom resource to k8s namespace `submarine-user-test`, can now only be connected to with host `submarine-user-test.submarine` as the destination e.g. via URL `http://submarine-user-test.submarine`.

To configure this, edit the newly added `spec.virtualservice.hosts` field of the submarine CR and `spec.host` of the virtual service will be set accordingly. For consistency, `spec.gateways` of the virtual service can now also be set through editing the submarine CR.

### What type of PR is it?
Feature

### Todos
* [x] - update operator source file and artifact file
* [x] - add e2e test cases
* [x] - add new development doc

### What is the Jira issue?

https://issues.apache.org/jira/browse/SUBMARINE-976

Follow submarine-cloud-v3/docs/developer-guide.md

### Questions:


submarine-cloud-v3/controllers/submarine_virtualservice.go
```diff
@@ -39,6 +39,7 @@ func (r *SubmarineReconciler) newSubmarineVirtualService(ctx context.Context, su
 		r.Log.Error(err, "ParseVirtualService")
 	}
 	virtualService.Namespace = submarine.Namespace
+	virtualService.Spec.Hosts = append(virtualService.Spec.Hosts, submarine.Namespace+".submarine")
```
cdmikechen, line 42: Is this `hosts` a necessary option? 
If the user has deployed an ingress controller (e.g. nginx), my understanding is that it should be possible for the user to define the hostname themselves and configure some https certificate at the same time.

joshvictor1024: I think this only acts as a default value. Users can re-define their hostname using `kubectl edit` after the virtualservice is created. Is this ok?

cdmikechen: @joshvictor1024 
In that case, is it possible for us to add this configuration to the `Custom Resource`? 
Since `VirtualService` by istio is created by submarine operator, I think we should also have the ability to configure it in CR.

joshvictor1024: @cdmikechen This makes sense. I will add field `host` to the submarine CRD. For a working default value (anything but `"*"`) to be assigned when creating the CR in an arbitrary namespace, this field will be optional and default to `<namespace>.submarine`. When a value is given, then that value overwrites the default one.

cdmikechen:
@joshvictor1024 
Hi~ I fixed the check failure on the `submarine-server`. If you have time, can you resubmit to see if the error is still there?
In the meantime, I would like to confirm if the changes in CR we can have in the test cases?

joshvictor1024:
@cdmikechen
The check on `submarine-server` is passing now, thank you. For the changes in the CR, right now there are no test cases for them. I'll change the test cases to check whether the value of `host` of the virtual service is correct, for when `spec.virtualservice.host` is set or not set in submarine CR.

cdmikechen:
LGTM 
@pingsutw Can you help continue to review the code?

xunliu:
LGTM

PR #987, opened 2022-08-12
SUBMARINE-1305. Fix submarine-server-core test error

cdmikechen:
### What is this PR for?
Currently, the test case class `SubmarineAuthSimpleTest` in `submarine-server-core` does not reset `submarine.auth.type` to `none` at the end (@After), which will cause subsequent test cases to be blocked.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Reset `submarine.auth.type` to `none` after test passed.

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1305

Have fixed.

No

### Questions:


cdmikechen:
@pingsutw 
Hi~ After PR https://github.com/apache/submarine/pull/979 has been merged, I found that the test cases in `submarine-server-core` would get blocked, probably due to a problem with my legacy test case (it didn't seem to be a problem in the test submitted at the time, I suspect it was caused by PRs that were merged after the PR was raised). I have now fixed the problem and would appreciate your help in reviewing it and merging it as soon as possible so that it does not affect other PRs.

pingsutw (approved):
@cdmikechen good catch, thank you

PR #989, opened 2022-08-15
SUBMARINE-1283. copy data for experiment before it running via distcp to minio

FatalLin:
### What is this PR for?
This is a prototype of experiment prehandler, once the required arguments has been set, submarine would put an init-container on the main pod. The init container would copy the source data to minio to the path /submarine/${experimentId}.
note: the init container would be add under:
TFJob: ps
PytorchJob: master
XGboostJob: master

### What type of PR is it?
Feature

### Todos
add a housekeeping container to clean the copied data.

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1283
Should add another test for it.
### Questions:


submarine-experiment-prehandler/fs_prehandler/hdfs_prehandler.py
```diff
@@ -15,26 +15,39 @@
...
         logging.info('DEST_PATH:%s' % self.dest_path)
         logging.info('ENABLE_KERBEROS:%s' % self.enable_kerberos)
 
-        self.fs = HadoopFileSystem(host=self.hdfs_host, port=self.hdfs_port)
-
     def process(self):
-        self.fs.get(self.hdfs_source, self.dest_path, recursive=True)
+
+        p = subprocess.run([self.hadoop_home+'/bin/hadoop', 'distcp'
+            , '-Dfs.s3a.endpoint=http://' + self.dest_minio_host + ':' + self.dest_minio_port + '/'
```
cdmikechen, line 47: I have tested it and found that we need to add `-Dfs.s3a.path.style.access=true`, otherwise the map step will be stuck at 4%

pingsutw: related to https://issues.apache.org/jira/browse/HADOOP-12963

.github/scripts/build-image-locally-v3.sh
```diff
@@ -17,12 +17,14 @@
 #
 
 SUBMARINE_VERSION="0.8.0-SNAPSHOT"
```
cdmikechen, line 19: I am consulting on an unrelated matter:
Can we get our version information directly from the `pom.xml` so that we don't need to do a lot of text replacement and version number validation each time we make a change to the version? For example
```shell
awk '/<\/*version\/*>/{gsub(/[[:space:]]*<\/*version\/*>/,"");print $0}' pom.xml | head -n 1
```

cc @pingsutw @xunliu

submarine-experiment-prehandler/fs_prehandler/hdfs_prehandler.py
```diff
@@ -15,26 +15,43 @@
```
cdmikechen, line 15: I tried to build the docker image and found that roughly this image has 2.8G in local. 
I don't know if this will affect the overall user experience, after all, if the image is too large, it will take too much time to pull it. 
Or is it possible to reduce it to a simple java code, so that python and the whole hadoop will not be installed , and the image will contain only the jars needed inside.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentSpec.java
```diff
@@ -63,6 +64,14 @@ public CodeSpec getCode() {
...
     this.code = code;
   }
+  
+  public Map<String, String> getExperimentHandlerSpec() {
+    return experimentHandlerSpec;
+  }
+
+  public void setExperimentHandlerSpec(Map<String, String> experimentHandlerSpec) {
+    this.experimentHandlerSpec = experimentHandlerSpec;
+  }
```
cdmikechen, line 74: I think we could add some more test cases to the `submarine-server/server-submitter/submitter-k8s/src/test/java/org/apache/submarine/server/submitter/k8s/ExperimentSpecParserTest.java` to pass the codecov checks.

.github/scripts/build-image-locally-v3.sh
FatalLin: In my opinion, maybe we could create another text file call submarine.version or something else and it would be the source of project version, all the metadata could read from it, include pom.xml, build script, and anything else.

submarine-experiment-prehandler/fs_prehandler/hdfs_prehandler.py
FatalLin: that's a good point, I take the whole hadoop deployment in it because we could access hdfs and s3 in the same approach, I could check is there any lighter way to execute distcp.

submarine-server/server-api/src/main/java/org/apache/submarine/server/api/spec/ExperimentSpec.java
FatalLin: good call.

.github/scripts/build-image-locally-v3.sh
cdmikechen: @FatalLin 
Created a new issue https://issues.apache.org/jira/browse/SUBMARINE-1328
I will follow up on this issue further.

cdmikechen:
@FatalLin 
After https://github.com/apache/submarine/pull/994 have been merged, there is a flie confilict. After the conflict is resolved, I think this PR can be merged first.

Meanwhile I reconsidered later today after the meeting and I think it should be possible to adapt this prehandler operation by adding a `load_dataset` method to the `submarine-sdk`.
For example, we could modify quickstart to look like this:
```python
hdfs_config = {'dfs.nameservices': 'example-cluster', 'dfs.ha.namenodes.example-cluster': 'nn1,nn2', ...}
dataset = submarine.load_datasets('hdfs', hdfs_config, 'hdfs://warehouse/datasets/***.parquet')
```

The underlying implementation of `load_dataset` registers the prehandler service as pod, and performs the dataset loading\training of the experiment after successful data copying. In distributed mode, we can do blocking before the pod is finished, so that each worker can wait for the data copying to complete.
I will follow up later to see how `kubeflow` does distributed dataset loading. In the meantime, there are some good ideas from another project [huggingface-datasets](https://github.com/huggingface/datasets) that I think we should learn from (huggingface also seems to download datasets locally first).

submarine-experiment-prehandler/fs_prehandler/hdfs_prehandler.py
cdmikechen: large images size issues to be fixed in subsequent issues

cdmikechen (approved):
LGTM.
@FatalLin I've changed the sample code for `load_dataset` design, do you think this will solve some of our common dataset problems?

FatalLin:
> LGTM. @FatalLin I've changed the sample code for `load_dataset` design, do you think this will solve some of our common dataset problems?

I'm wondering how will you do to block the pods?

cdmikechen:
@FatalLin 
I think there should be 2 ways to do this:
1. the pod's status has different values in different cases, and we can block the program processing based on the status. If the status of the pod is successful, the program will be executed backwards, and if the pod fails, an exception can be thrown.
```yaml
status:
  phase: Running
---
status:
  phase: Failed
---
status:
  phase: Succeeded
```

2. synchronise the execution of the prehandler to the database and determine if it is blocked based on the current status of the specified prehandler.

I prefer the second approach, given that we may need to probably be tracking the dataset detail during experiment and serving. We can discuss the details in more detail in the next regular meeting.
cc @pingsutw

FatalLin:
> @FatalLin I think there should be 2 ways to do this:
> 
> 1. the pod's status has different values in different cases, and we can block the program processing based on the status. If the status of the pod is successful, the program will be executed backwards, and if the pod fails, an exception can be thrown.
> 
> ```yaml
> status:
>   phase: Running
> ---
> status:
>   phase: Failed
> ---
> status:
>   phase: Succeeded
> ```
> 
> 2. synchronise the execution of the prehandler to the database and determine if it is blocked based on the current status of the specified prehandler.
> 
> I prefer the second approach, given that we may need to probably be tracking the dataset detail during experiment and serving. We can discuss the details in more detail in the next regular meeting. Right now I will merge it. cc @pingsutw

I mean WHERE you want to add the blockers? another initContainer for each experiment pods?

cdmikechen:
> > @FatalLin I think there should be 2 ways to do this:
> > 
> > 1. the pod's status has different values in different cases, and we can block the program processing based on the status. If the status of the pod is successful, the program will be executed backwards, and if the pod fails, an exception can be thrown.
> > 
> > ```yaml
> > status:
> >   phase: Running
> > ---
> > status:
> >   phase: Failed
> > ---
> > status:
> >   phase: Succeeded
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. synchronise the execution of the prehandler to the database and determine if it is blocked based on the current status of the specified prehandler.
> > 
> > I prefer the second approach, given that we may need to probably be tracking the dataset detail during experiment and serving. We can discuss the details in more detail in the next regular meeting. Right now I will merge it. cc @pingsutw
> 
> I mean WHERE you want to add the blockers? another initContainer for each experiment pods?

My assumption is that the implementation of the blocking logic is not done at the k8s level, but should be done in the way that `submarine-sdk` handles this.
Assuming we use the second option, the first worker will create the prehandler's pod resource and add/update the status of the dataset row in database to `pulling`. the second worker will first check if the current request to pull the data is executing, and if it finds the dataset is pulling then block the process and wait for the task to finish.

FatalLin:
> > > @FatalLin I think there should be 2 ways to do this:
> > > 
> > > 1. the pod's status has different values in different cases, and we can block the program processing based on the status. If the status of the pod is successful, the program will be executed backwards, and if the pod fails, an exception can be thrown.
> > > 
> > > ```yaml
> > > status:
> > >   phase: Running
> > > ---
> > > status:
> > >   phase: Failed
> > > ---
> > > status:
> > >   phase: Succeeded
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 2. synchronise the execution of the prehandler to the database and determine if it is blocked based on the current status of the specified prehandler.
> > > 
> > > I prefer the second approach, given that we may need to probably be tracking the dataset detail during experiment and serving. We can discuss the details in more detail in the next regular meeting. Right now I will merge it. cc @pingsutw
> > 
> > 
> > I mean WHERE you want to add the blockers? another initContainer for each experiment pods?
> 
> My assumption is that the implementation of the blocking logic is not done at the k8s level, but should be done in the way that `submarine-sdk` handles this. Assuming we use the second option, the first worker will create the prehandler's pod resource and add/update the status of the dataset row in database to `pulling`. the second worker will first check if the current request to pull the data is executing, and if it finds the dataset is pulling then block the process and wait for the task to finish.

I see, I think its an interesting design, but there's a tradeoff that once we combine the feature with our sdk, the user's image 
 is no longer portable for other environment. For example, they might only leverage submarine for developing, and they have some standalone  production environment.

cdmikechen:
@FatalLin 
Yes, loading datasets is indeed a method based on the need to adapt for different environments, whether it is in a cloud environment is something that can be determined within the sdk via the k8s api. If it's in the cloud then you can register a prehandler pod resource for doing data synchronisation to minio/s3, or pull the data locally for processing. Users should be able to choose their own method of pulling data that is appropriate for their current environment.
I think more details can be added in `submarine.load_datasets`, and I don't have much event confirmation on the huggingface project yet. I think it might also give us a new way of thinking.

PR #992, opened 2022-08-28
SUBMARINE-1315. Fix model can not serve error

cdmikechen:
### What is this PR for?
The model serve now registers the wrong namespace for resources, and taking the model name directly as a resource name can lead to problems with illegal characters.
image

### What type of PR is it?
Bug Fix

### Todos
* [x] - Model serve codes refactoring.
* [x] - Replace default to submarine-server namespace.
* [x] - Add a primary key id for `registered_model`(model_id) and `model_version` (model_version_id)
* [x] - Replace model service name to `submarine-model-${id}-@{modelId}` in k8s resource metadata name. 
* [x] - Replace SeldonGraph container name to `version-${modelVersion}`.
* [x] - Replace ingress routing to `/seldon/${namespace}/${modelVersionId}/${modelVersion}/` and add owner reference for cascaded deletion of operator.
* [x] - Add save_model method and serve predictions example in example quick-start .

### What is the Jira issue?
https://issues.apache.org/jira/projects/SUBMARINE/issues/SUBMARINE-1315

Add a simple test case to test k8s resource.

There is still a legacy of problems. Now that seldon automatically creates the serve services (http and rpc) using istio, the `VirtualService` we create for our own serve will be duplicated, and the url provided by the seldon returned by calling swagger with our own `VirtualService` (like http://localhost:32080/seldon/submarine-user-test/1/1/api/v1.0/doc/) is still seldon http `VirtualService` url.

The first was created by us and the following two by seldon. Both are currently accessible.
image

Swagger UI
image

### Questions:

### DIFF
--- a/dev-support/database/submarine-model.sql
+++ b/dev-support/database/submarine-model.sql
@@ -11,6 +11,8 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
+-- Since the associated tables have primary and foreign key constraints,
+-- we need to delete them all before creating them
 DROP TABLE IF EXISTS `param`;
 DROP TABLE IF EXISTS `metric`;
 DROP TABLE IF EXISTS `model_version_tag`;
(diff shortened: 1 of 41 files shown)


pingsutw (approved):

cdmikechen:
@pingsutw 
Hi~ There are still some seldon issues that need to do and I expect to do these issues this weekend. You could be able to start the review by the time I get the WIP removed.
Thanks ~

cdmikechen:
@pingsutw @KUAN-HSUN-LI Hi~ 
I have fixed all the problems I have encountered with serving.
However, there is still a problem: Seldon will also use istio to create the corresponding `Virtualservice`. I have uploaded the relevant screenshots and written contents to the PR description document. If you are free, could you review it?

xunliu:
LGTM

PR #995, opened 2022-09-12
SUBMARINE-1324. Fix the experiment pods label selector after using the new training operator

cdmikechen:
### What is this PR for?
After updated to new training operator, the metrics/logs/params of the experiment cannot be shown, which is found that the pod label had changed to `job-name`.

### What type of PR is it?
Bug Fix

### Todos
* [x] - Change label to job-name

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1324

Consider adding after refactoring.

image

### Questions:


cdmikechen:
This issue has been merged for the time being as it will affect the flow of the experiment.

PR #999, opened 2022-09-28
SUBMARINE-1293. Move seldon-core install to helm

hhcs9527:
### What is this PR for?
Move seldon-core install to helm

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1293

Make sure we can install the sekdon-core correctly.

### Questions:


helm-charts/submarine/templates/seldon-secret.yaml
```diff
@@ -1,3 +1,4 @@
+
```
cdmikechen, line 1: I've checked this file and found that we need to replace namespace:
```yaml
namespace: {{ .Release.Namespace }}
```
And please ensure that the license information starts on the first line.

submarine-serve/installation/install.sh
```diff
@@ -33,13 +33,3 @@ if [[ ! -f ${CURRENT_PATH}/istioctl ]]; then
...
 
 ${CURRENT_PATH}/istioctl install --skip-confirmation
-
-kubectl create ns seldon-system
-kubectl apply -f ${CURRENT_PATH}/seldon-secret.yaml
-helm install seldon-core seldon-core-operator \
-    --repo https://storage.googleapis.com/seldon-charts \
-    --namespace seldon-system \
-    --version 1.10.0 \
-    --set istio.enabled=true \
```
cdmikechen, line 43: We also need to synchronise the variables(`istio.enabled=true` and `executor.defaultEnvSecretRefName=submarine-serve-secret 2> /dev/null`) declared here with `values.yaml` to ensure consistency of configuration.

cdmikechen:
Hi~ @hhcs9527 
Thanks your contribution. I've left some comments that may need to be modified and I'll verify the deployment later.

cdmikechen (changes requested):

helm-charts/submarine/Chart.yaml
```diff
@@ -23,6 +23,9 @@ version: 0.8.0-SNAPSHOT
 icon: https://submarine.apache.org/assets/themes/submarine/img/submarine_white_logo.png
 dependencies:
   - name: training-operator
-    version: "0.1.0"
+    version: "0.7.0"
   - name: notebook-controller
     version: "0.1.0"
+  - name: seldon-core-operator
+    version: "1.10.0"
+    repository: https://storage.googleapis.com/seldon-charts
```
cdmikechen, line 31: If the user already has installed `seldon`, I think we can add a condition to make helm not install this dependency. For example:
```yaml
condition: seldon-core-operator.enabled
```

cdmikechen (changes requested):

helm-charts/submarine/templates/seldon-secret.yaml
cdmikechen: I have since discovered that deployments based on this file currently have another problem that needs to be solved (https://issues.apache.org/jira/browse/SUBMARINE-1332). I would therefore suggest that we can remove this file for the time being.

hhcs9527: ok, I got it.

submarine-serve/installation/install.sh
hhcs9527: Thanks for the reminding.

helm-charts/submarine/values.yaml
```diff
@@ -45,3 +45,11 @@ podSecurityPolicy:
 istio:
   enabled: true
   gatewaySelector: ingressgateway
+
```
cdmikechen, line 48: Could you verify that this configuration works? I remember that the configuration of the dependency chart should have name as the root node, e.g:
```yaml
seldon-core-operator:
  executor:
    defaultEnvSecretRefName: submarine-serve-secret 2> /dev/null
  istio:
    enabled: true
``

**Review by cdmikechen — CHANGES_REQUESTED** [review]

**Review by cdmikechen — CHANGES_REQUESTED** [review]
@hhcs9527 
Hi. What I thought was that we did not need to deal with `submarine-serve/installation/seldon-secret.yaml` for the time being.
Also, the namespace of the gateway needs to be the same as the helm install namespace.

I'll turn on github workflow when this part is fixed, there may still be issues to resolve when running test cases, and I will help you with this part when the time comes.

**Comment by hhcs9527** [comment]
> @hhcs9527 Hi. What I thought was that we did not need to deal with `submarine-serve/installation/seldon-secret.yaml` for the time being. Also, the namespace of the gateway needs to be the same as the helm install namespace.
> 
> I'll turn on github workflow when this part is fixed, there may still be issues to resolve when running test cases, and I will help you with this part when the time comes.

I think I misunderstand your meaning. I am sorry.
Many thanks for your help on this.

File: helm-charts/submarine/templates/seldon-gateway.yaml [hunk 5]
```diff
@@ -18,7 +18,7 @@ apiVersion: networking.istio.io/v1alpha3
 kind: Gateway
 metadata:
   name: seldon-gateway
-  namespace: istio-system
+  namespace: default
```
cdmikechen, line 21: please use  {{ .Release.Namespace }}

helm-charts/submarine/Chart.yaml
```diff
@@ -23,6 +23,10 @@ version: 0.8.0-SNAPSHOT
 icon: https://submarine.apache.org/assets/themes/submarine/img/submarine_white_logo.png
 dependencies:
   - name: training-operator
-    version: "0.1.0"
+    version: "0.7.0"
   - name: notebook-controller
     version: "0.1.0"
+  - name: seldon-core-operator
```
cdmikechen, line 29: I think we should add the `helm dep update` to https://github.com/apache/submarine/blob/master/website/docs/gettingStarted/quickstart.md and add the dependency description.

helm-charts/submarine/templates/seldon-secret.yaml
```diff
@@ -18,7 +18,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: submarine-serve-secret
-  namespace: submarine-user-test
+  namespace: default
```
cdmikechen, line 21: we should not move and changed this file, please.

cdmikechen (changes requested):

cdmikechen (approved):
LGTM
Thanks for your efforts on this PR

PR #1016, opened 2022-11-15
SUBMARINE-1348. Fix implicit Optional check style issues for mypy

JackLiu00521:
### What is this PR for?
Fix mypy python check error (cause by mypy recent change: --no-implicit-optional by default)
https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html

### What type of PR is it?
[Bug Fix]

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1348

Python check style test will now pass

### Questions:

### DIFF
--- a/dev-support/style-check/python/lint.sh
+++ b/dev-support/style-check/python/lint.sh
@@ -21,7 +21,7 @@ cd "$FWDIR"
 cd ../../../
 
 # Check imports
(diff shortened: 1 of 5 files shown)


cdmikechen (approved):
LGTM, thanks for fixing this issue~

PR #1018, opened 2022-11-19
SUBMARINE-1347. Fix the syntax error reported in sonarcloud.

hhcs9527:
### What is this PR for?
Fix the syntax error reported in sonarcloud.

### What type of PR is it?
Bug Fix 

### Todos
* [ ] - Task

### What is the Jira issue?
[SUBMARINE-1347](https://issues.apache.org/jira/browse/SUBMARINE-1347)
### Questions:


hhcs9527:
@cdmikechen please take a look at the pr, thanks.

pingsutw (approved):

cdmikechen:
@pingsutw @hhcs9527 
I think there is still a problem with this change in `XGBoostJobList`, as there is only a `get` method, so we need to initialize the relevant variables, otherwise there will be a NPE.
image

PR #1020, opened 2022-11-22
SUBMARINE-1349. Fix the syntax error reported in sonarcloud and add init value for apiversion in XGboostjobList.java

hhcs9527:
### What is this PR for?

1. Fix the syntax error reported in sonarcloud
2. Add init value for apiversion in XGboostjobList.java
3. change the following class to a singleton, [ModelVersionService.java](https://github.com/apache/submarine/pull/1020/files#diff-0b498de1dff04b42428e969cc3e0fb1376b75b9f33db4ae8d9740c7e1a2e43be), [RegisteredModelService.java](https://github.com/apache/submarine/pull/1020/files#diff-5894fdc7eea50e5065c43846308e16847562089db57b207ea23ceef86c639668),[RegisteredModelTagService.java](https://github.com/apache/submarine/pull/1020/files#diff-56494d2d82488701eb8e3738e8ff3835a27e7f5add8c9d3421ee35e4e2b09661), and the managers used thems.

### What type of PR is it?
Bug Fix

### Todos
* [ ] - Task

### What is the Jira issue?
[SUBMARINE-1349](https://issues.apache.org/jira/browse/SUBMARINE-1349)
### Questions:

### DIFF
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
@@ -50,12 +50,13 @@ public class ModelVersionManager {
    *
    * @return object
    */
-  public static synchronized ModelVersionManager getInstance() {
-    if (manager == null) {
-      manager = new ModelVersionManager(new ModelVersionService(), new ModelVersionTagService(),
-          new Client());
-    }
-    return manager;
+  private static class ModelVersionManagerHolder {
+    private static ModelVersionManager manager = new ModelVersionManager(ModelVersionService.getInstance(), new ModelVersionTagService(),
+    Client.getInstance());
+  }
+
+  public static ModelVersionManager getInstance() {
+    return ModelVersionManager.ModelVersionManagerHolder.manager;
   }
 
   @VisibleForTesting
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
@@ -51,8 +51,9 @@ public class ModelVersionManager {
    * @return object
    */
   private static class ModelVersionManagerHolder {
-    private static ModelVersionManager manager = new ModelVersionManager(ModelVersionService.getInstance(), new ModelVersionTagService(),
-    Client.getInstance());
+    private static ModelVersionManager manager = new ModelVersionManager(ModelVersionService.getInstance(),
+                                                                        new ModelVersionTagService(),
+                                                                        Client.getInstance());
   }
 
   public static ModelVersionManager getInstance() {
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
@@ -50,14 +50,10 @@ public class ModelVersionManager {
    *
    * @return object
    */
-  public static ModelVersionManager getInstance() {
+  public static synchronized ModelVersionManager getInstance() {
     if (manager == null) {
-      synchronized (ModelVersionManager.class) {
-        if (manager == null) {
-          manager = new ModelVersionManager(new ModelVersionService(), new ModelVersionTagService(),
-            new Client());
-        }
-      }
+      manager = new ModelVersionManager(new ModelVersionService(), new ModelVersionTagService(),
+          new Client());
     }
     return manager;
   }
--- a/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/RegisteredModelManager.java
+++ b/submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/RegisteredModelManager.java
@@ -37,7 +37,6 @@ import org.apache.submarine.server.s3.Client;
  * Registered model manager.
  */
 public class RegisteredModelManager {
-  private static RegisteredModelManager manager;
   /* Registered model service */
   private final RegisteredModelService registeredModelService;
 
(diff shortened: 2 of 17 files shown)


hhcs9527:
Hi @cdmikechen , Please take a look at this PR, thanks.

submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
```diff
@@ -50,14 +50,10 @@ public class ModelVersionManager {
    *
    * @return object
    */
-  public static ModelVersionManager getInstance() {
+  public static synchronized ModelVersionManager getInstance() {
```
cdmikechen, line 53: We should not change the position of `synchronized`, changing it on this will cause the whole method to be locked at the time of the call.

submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/RegisteredModelManager.java
```diff
@@ -54,14 +54,10 @@ public class RegisteredModelManager {
    *
    * @return object
    */
-  public static RegisteredModelManager getInstance() {
+  public static synchronized RegisteredModelManager getInstance() {
```
cdmikechen, line 57: Same as above

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/model/xgboostjob/XGBoostJobList.java
```diff
@@ -27,7 +27,7 @@
 
 public class XGBoostJobList implements KubernetesListObject {
   @SerializedName("apiVersion")
-  private String apiVersion;
+  private String apiVersion = XGBoostJob.CRD_XGBOOST_API_VERSION_V1;
 
   @SerializedName("kind")
   private String kind;
```
cdmikechen, line 33: Does `kind` have the same problem?

cdmikechen (changes requested):

submarine-server/server-submitter/submitter-k8s/src/main/java/org/apache/submarine/server/submitter/k8s/model/xgboostjob/XGBoostJobList.java
hhcs9527: @cdmikechen kind doesn't show the problem in sonarcloud.

hhcs9527:
@cdmikechen just updated with some modifications, please take a look. (singleton, and their reference)

cdmikechen:
@hhcs9527 
There are some check style and build task errors, please take care of this part first, thanks~

submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
```diff
@@ -43,19 +45,38 @@
 
 public class Client {
   public MinioClient minioClient;
+  public static Map<String,MinioClient> minioClientFactory = new HashMap<String, MinioClient>();
+  public static Map<String,Client> clientFactory = new HashMap<String, Client>();
+
+  public static Client getClient(String endpoint) {
```
cdmikechen, line 51: Maybe you can use a `enum` class to replace inner factory class to avoid client double checks.

hhcs9527: Hi @cdmikechen , I agree with you.
But the enum type is more suitable for the predefined class (no argument to init). In our case, we need to init different users in testing, which cannot be predefined; that's why I use double-check locking.

cdmikechen: Java enum class can be used inside a class to create a constructor with parameters and as a singleton. Variables are only one of its basic uses.

hhcs9527:
@cdmikechen just updated Java enum class, please take a look

submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
```diff
@@ -50,16 +50,14 @@ public class ModelVersionManager {
    *
    * @return object
    */
+  private static class ModelVersionManagerHolder {
+    private static ModelVersionManager manager = new ModelVersionManager(ModelVersionService.getInstance(),
+                                                                        new ModelVersionTagService(),
```
cdmikechen, line 55: new class param should also be avoided here.

submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
```diff
@@ -45,30 +47,57 @@
 /**
  * S3(Minio) default client
  */
-public class Client {
+public enum Client {
+  DEFAULT(SubmarineConfiguration.getInstance().getString(SubmarineConfVars.ConfVars.SUBMARINE_S3_ENDPOINT)),
```
cdmikechen, line 51: I don't think it's necessary to add an endpoint attribute to the enum, which can't declare a variable number of endpoints and has to use the internal map to handle it.
```java
public enum Client {
  INSTANCE;
  public static Client getInstance(String endpoint) {
    // ....
  }
}

Client.INSTANCE.getInstance("http://localhost:9000");
```

cdmikechen (changes requested):
codecov currently finds that the current changes do not have some test cases, please add some test cases for the minio client section. The other sections can be left alone for now.

submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
hhcs9527: Hi @cdmikechen ,
Not sure what you mean by adding some test cases for the minio client section here.

submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
hhcs9527: Hi @cdmikechen ,
Not sure what you mean here as well. Are you suggesting not initiating the enum variable?

submarine-server/server-core/src/main/java/org/apache/submarine/server/manager/ModelVersionManager.java
cdmikechen: New code changes need to be supported by test cases, otherwise they will reduce code coverage and fail to prove code robustness.

submarine-server/server-core/src/main/java/org/apache/submarine/server/s3/Client.java
cdmikechen: When using an enum as a factory to get a single instance object, using a limited number of enum definitions does not achieve the final effect of a dynamic factory

cdmikechen (changes requested):
Hi~ There are still some test case failed and we can merge when you have resolved this issue.

hhcs9527:
@cdmikechen we still facing the same error for being unable to connect to the local host.
Any good suggestion?

cdmikechen (approved):
LGTM
I will merge this pr after python cicd fixed

PR #1029, opened 2022-12-06
SUBMARINE-1358. fix the potential timeout issue while building jupyter image

FatalLin:
### What is this PR for?
The github action hit the timeout exception while building the image of jupyter and jupyter-gpu time to time  since the previous days. After couple testing, I found it indeed takes infinite time to install CVXPY with conda. Per the discussion on weekly meeting, I change the way to install CVXPY from conda to pip, and it also bring s some side-effect here like the some dependent package is too old and cause "use_2to3 is invalid" error, so some of the package had also been upgraded as well in this PR.

### What type of PR is it?
Improvement

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1358?
I build and deploy the image in my local environment and looks like everything is worked.
### Questions:

### DIFF
--- a/dev-support/docker-images/jupyter-gpu/Dockerfile
+++ b/dev-support/docker-images/jupyter-gpu/Dockerfile
@@ -118,7 +118,7 @@ RUN git clone --depth=1 https://github.com/apache/submarine && \
     rm submarine -rf
 
 # Add qlib example in notebook
-RUN wget https://raw.githubusercontent.com/microsoft/qlib/main/examples/workflow_by_code.ipynb -P $HOME
+RUN wget https://raw.githubusercontent.com/microsoft/qlib/v0.9.0/examples/workflow_by_code.ipynb -P $HOME
 
 
 EXPOSE $NB_PORT
--- a/dev-support/docker-images/jupyter-gpu/Dockerfile
+++ b/dev-support/docker-images/jupyter-gpu/Dockerfile
@@ -93,8 +93,8 @@ RUN wget --quiet https://repo.continuum.io/miniconda/Miniconda3-py37_${MINICONDA
     if [ ! $PYTHON_VERSION = 'default' ]; then conda install --freeze-installed --yes python=$PYTHON_VERSION; fi && \
     conda list python | grep '^python ' | tr -s ' ' | cut -d '.' -f 1,2 | sed 's/$/.*/' >> $CONDA_DIR/conda-meta/pinned && \
     # nodejs is used for jupyter lab build
-    conda install --freeze-installed -c conda-forge nodejs==14.20.1 jupyterlab jupyterlab-git && \
-    pip install cvxpy && \
+    conda install --freeze-installed -c conda-forge nodejs==14.20.1 jupyterlab==3.5.2 jupyterlab-git==0.41.0 && \
+    pip install cvxpy==1.2.2 && \
     jupyter lab build && \
     # remove node_modules
     rm -rf /home/$NB_USER/.cache/yarn && \
@@ -105,7 +105,7 @@ RUN wget --quiet https://repo.continuum.io/miniconda/Miniconda3-py37_${MINICONDA
     conda clean -a -y
 
 # Install latest stable qlib
-RUN pip --no-cache-dir install pyqlib
+RUN pip --no-cache-dir install pyqlib==0.8.6
 
 # Install latest sumbarine python sdk and notebook
 RUN git clone --depth=1 https://github.com/apache/submarine && \
--- a/dev-support/docker-images/jupyter-gpu/Dockerfile
+++ b/dev-support/docker-images/jupyter-gpu/Dockerfile
@@ -93,7 +93,8 @@ RUN wget --quiet https://repo.continuum.io/miniconda/Miniconda3-py37_${MINICONDA
     if [ ! $PYTHON_VERSION = 'default' ]; then conda install --freeze-installed --yes python=$PYTHON_VERSION; fi && \
     conda list python | grep '^python ' | tr -s ' ' | cut -d '.' -f 1,2 | sed 's/$/.*/' >> $CONDA_DIR/conda-meta/pinned && \
     # nodejs is used for jupyter lab build
-    conda install --freeze-installed -c conda-forge nodejs jupyterlab jupyterlab-git cvxpy==1.0.21 && \
+    conda install --freeze-installed -c conda-forge nodejs==14.20.1 jupyterlab jupyterlab-git && \
+    pip install cvxpy && \
     jupyter lab build && \
     # remove node_modules
     rm -rf /home/$NB_USER/.cache/yarn && \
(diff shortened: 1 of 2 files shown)


cdmikechen:
@FatalLin 
Hi~ I would like to know if this fix is already available for building images locally, please?

FatalLin:
> @FatalLin Hi~ I would like to know if this fix is already available for building images locally, please?

Yes, just like I mentioned in description, both the jupyter and jupyter-gpu could be built in my local environment.

cdmikechen (commented):
@FatalLin I think this should be fine, but I just have a question: 
Can we add a version range limit on the changes like @pingsutw  mentioned last time? This would avoid the problem of new versions causing build failures again in the future.

FatalLin:
> @FatalLin I think this should be fine, but I just have a question: Can we add a version range limit on the changes like @pingsutw mentioned last time? This would avoid the problem of new versions causing build failures again in the future.

do you mean fix the version of all dependency?

cdmikechen:
> > @FatalLin I think this should be fine, but I just have a question: Can we add a version range limit on the changes like @pingsutw mentioned last time? This would avoid the problem of new versions causing build failures again in the future.
> 
> do you mean fix the version of all dependency?

I mean the lines you changed with `pip install`.

FatalLin:
> Contributor



> > > > > @FatalLin I think this should be fine, but I just have a question: Can we add a version range limit on the changes like @pingsutw mentioned last time? This would avoid the problem of new versions causing build failures again in the future.
> > > > 
> > > > 
> > > > do you mean fix the version of all dependency?
> > > 
> > > 
> > > I mean the lines you changed with `pip install`.
> > 
> > 
> > I think we could fix the version at 1.2.2 - which would be the same with the current command, Im not favor in using a range of version on single dependency is because somehow it may bring unexpected changes from their own dependency, just like it require the version of nodeJs is higher than 14.0 this time.
> 
> Is it possible to determine the range/max of versions of `cvxpy` and `pyqlib` or `nodejs` supported by conda 4.11 in the official jupyter project? Or is it possible that using the conda command will automatically install the best version for the current environment?
> 
> If we don't specify the version range, every install command will get the latest version and at some point in the future if there is a conflict between the new versions it will again break the build, which is what I'm worried about.

I don't case, if your purpose is trying to avoid potential conflict, why dont we just fix all the version here?

cdmikechen:
> > Contributor
> 
> > > > > > @FatalLin I think this should be fine, but I just have a question: Can we add a version range limit on the changes like @pingsutw mentioned last time? This would avoid the problem of new versions causing build failures again in the future.
> > > > > 
> > > > > 
> > > > > do you mean fix the version of all dependency?
> > > > 
> > > > 
> > > > I mean the lines you changed with `pip install`.
> > > 
> > > 
> > > I think we could fix the version at 1.2.2 - which would be the same with the current command, Im not favor in using a range of version on single dependency is because somehow it may bring unexpected changes from their own dependency, just like it require the version of nodeJs is higher than 14.0 this time.
> > 
> > 
> > Is it possible to determine the range/max of versions of `cvxpy` and `pyqlib` or `nodejs` supported by conda 4.11 in the official jupyter project? Or is it possible that using the conda command will automatically install the best version for the current environment?
> > If we don't specify the version range, every install command will get the latest version and at some point in the future if there is a conflict between the new versions it will again break the build, which is what I'm worried about.
> 
> I don't get it, if your purpose is trying to avoid potential conflict, why dont we just fix all the version here?

I'm not a heavy user of conda, but I think conda should install the version of the resource that matches the dependencies according to its own environment (4.11).
So I think we should just make sure that the `cvxpy` and `pyqlib` version which you changed using the `pip install` command matches or try using the `conda install` command to verify it.

FatalLin:
> > > Contributor
> > 
> > 
> > > > > > > @FatalLin I think this should be fine, but I just have a question: Can we add a version range limit on the changes like @pingsutw mentioned last time? This would avoid the problem of new versions causing build failures again in the future.
> > > > > > 
> > > > > > 
> > > > > > do you mean fix the version of all dependency?
> > > > > 
> > > > > 
> > > > > I mean the lines you changed with `pip install`.
> > > > 
> > > > 
> > > > I think we could fix the version at 1.2.2 - which would be the same with the current command, Im not favor in using a range of version on single dependency is because somehow it may bring unexpected changes from their own dependency, just like it require the version of nodeJs is higher than 14.0 this time.
> > > 
> > > 
> > > Is it possible to determine the range/max of versions of `cvxpy` and `pyqlib` or `nodejs` supported by conda 4.11 in the official jupyter project? Or is it possible that using the conda command will automatically install the best version for the current environment?
> > > If we don't specify the version range, every install command will get the latest version and at some point in the future if there is a conflict between the new versions it will again break the build, which is what I'm worried about.
> > 
> > 
> > I don't get it, if your purpose is trying to avoid potential conflict, why dont we just fix all the version here?
> 
> I'm not a heavy user of conda, but I think conda should install the version of the resource that matches the dependencies according to its own environment (4.11). So I think we should just make sure that the `cvxpy` and `pyqlib` version which you changed using the `pip install` command matches or try using the `conda install` command to verify it.

As my understanding, the version of the package we install via conda is more associated with the package itself(branch/tag/fix version...etc) and less with the conda's version. Since both the cvxpy and pyqlib is not required by the jupyter, and vise versa, I didn't find any official document to indicate the compatible version between these packages( it would be great appreciated if you point them out). If we want avoid the potential conflict in the future, based on my personal experience, the proper way is fixing all the package in a single version, no matter they are installed by conda or pip.

cdmikechen:
Here is the original issue link about `qlib`: https://issues.apache.org/jira/browse/SUBMARINE-725
We can discuss how to determine the version of `qlib` and the other related dependencies at the next meeting.

FatalLin:
@pingsutw could you please some suggestion on the version of these packages we installed? Is that still fulfill your origin thought if we fix all the version on which the current stable is? Or its better to keep them to install the version stable branch.

pingsutw (commented):
I was thinking that adding all the dependencies to the `requirements.in`, and use [pip-compile](https://github.com/jazzband/pip-tools) to generate a `requirements.txt` file with all the dependencies pinned. 
So we can just use `pip install -r requirements.txt` to install the package in the image, pip won't automatically update the package, break the build again

dev-support/docker-images/jupyter-gpu/Dockerfile
```diff
@@ -117,7 +118,7 @@ RUN git clone --depth=1 https://github.com/apache/submarine && \
     rm submarine -rf
 
 # Add qlib example in notebook
-RUN wget https://raw.githubusercontent.com/microsoft/qlib/main/examples/workflow_by_code.ipynb -P $HOME
+RUN wget https://raw.githubusercontent.com/microsoft/qlib/v0.9.0/examples/workflow_by_code.ipynb -P $HOME
```
cdmikechen, line 121: Is it possible to keep the version of qlib the same as the current file? I've tried and I think version 0.8.6 is also available for download.
https://raw.githubusercontent.com/microsoft/qlib/v0.8.6/examples/workflow_by_code.ipynb

Others LGTM

cdmikechen (approved):

PR #1044, opened 2023-02-12
SUBMARINE-1366. SQLAlchemy 2.0 upgrade causes compilation failure of pysubmarine

cdmikechen:
### What is this PR for?
After SQLAlchemy was upgraded to 2.0 some of the syntax was changed, which would cause syntax errors in the pysbumarine. We need to consider whether to keep the version at 1.4.x for now, or update the code and re-test it in light of the syntax issue. 
https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
Action run with exceptions:
https://github.com/apache/submarine/actions/runs/4096688603/jobs/7064519374

### What type of PR is it?
Bug Fix

### Todos
* [x] - Keep SQLAlchemy in 1.4.x

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1366

cicd

No

### Questions:
* Do the license files need updating? Yes

### DIFF
--- a/dev-support/style-check/python/mypy-requirements.txt
+++ b/dev-support/style-check/python/mypy-requirements.txt
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 mypy
-sqlalchemy[mypy]
+sqlalchemy[mypy]<2.0.0
 types-certifi
 types-python-dateutil
 types-PyYAML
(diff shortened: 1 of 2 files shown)


cdmikechen:
@pingsutw 
Hi~ Do you have time to look at this issue? My current solution is to limit it to version 1.4.x. Maybe we can adapt it to 2.0 when we get to 0.9.0.

pingsutw (approved):

pingsutw:
LGTM, thanks

PR #1072, opened 2023-06-22
SUBMARINE-1384. Fix some versions that were already wrong

cdmikechen:
### What is this PR for?
Fix some outdated versions

### What type of PR is it?
Documentation

### Todos
* [x] - Fix some versions and upgrade to right version

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1384

NA

### Questions:

### DIFF
--- a/website/docs/devDocs/Development.md
+++ b/website/docs/devDocs/Development.md
@@ -44,7 +44,7 @@ A brief introduction about these components:
 
 ### Prerequisites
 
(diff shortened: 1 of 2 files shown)


cdmikechen:
@pingsutw
Are there any other versions that need to be changed that you can add to the document when you check it?

pingsutw (approved):

PR #1092, opened 2023-07-23
SUBMARINE-1395. Remove unauthenticated login

cdmikechen:
### What is this PR for?
Backend server no longer supports login without authentication (except python-sdk), REST must authenticate first (/api/auth/token) and add a token in header to be accessed.

### What type of PR is it?
Feature 

### Todos
* [x] - Add `/api/auth/token` endpoint and sdk codes
* [x] - Remove none auth type
* [x] - Fix test cases  
* [x] - Reformat openapi.json by order fileds and tags (correct ordering will prevent each gen-sdk openapi.json having a different order)

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1395

CI test

NA

### Questions:


submarine-server/server-core/src/test/java/org/apache/submarine/server/AbstractSubmarineServerTest.java
```diff
@@ -237,10 +263,12 @@ protected static PostMethod httpPost(String path, String request, String mediaTy
     PostMethod postMethod = new PostMethod(URL + path);
     postMethod.setRequestBody(request);
     postMethod.setRequestHeader("Content-type", mediaType);
-    postMethod.getParams().setCookiePolicy(CookiePolicy.IGNORE_COOKIES);
+    //postMethod.getParams().setCookiePolicy(CookiePolicy.IGNORE_COOKIES);
```
pingsutw, line 266: Should we delete it?

cdmikechen: > Should we delete it?

Of course, with REST we should no longer explicitly support session/cookie later on.

pingsutw: SGTM

cdmikechen:
It looks like we meet `no space left on device` error again. 
https://github.com/apache/submarine/actions/runs/5693116661/job/15444228405?pr=1092

image

pingsutw (approved):

cdmikechen:
Currently the following error occurs inside the new ci testing process: 
https://github.com/apache/submarine/actions/runs/5840601129/job/15840001205
image

This is caused by some data within submarine_test that shouldn't be there. I'm currently looking for the exact cause.

cdmikechen:
Since some of the `submarine-server-core` tests are started based on the built-in jetty, that test cases do not go through the hard-coded logic of `org.apache.submarine.server.database.utils.checkCalledByTestMethod` at startup, making the startup of the test case go to submarine's database, causing a database processing exception.

https://github.com/apache/submarine/blob/f170037c1e7642c16cd5fa1707e85abd6252d839/submarine-server/server-database/src/main/java/org/apache/submarine/server/database/utils/MyBatisUtil.java#L83-L104

cdmikechen:
A related issue https://issues.apache.org/jira/browse/SUBMARINE-1297 has already been created before, and is currently being temporarily resolved by setting the environment variable for `JDBC_URL=jdbc:mysql://127.0.0.1:3306/submarine_test?useUnicode=true&characterEncoding=UTF-8&autoReconnect=true&allowMultiQueries=true&failOverReadOnly=false&zeroDateTimeBehavior=convertToNull&useSSL=false`. 
This should be fully resolved when the subsequent issue is resolved.

PR #1125, opened 2024-03-29
SUBMARINE-1417. Retrieve SUBMARINE_AUTH_SECRET from environment variable instead of using hard-coded value

laiyousin:
### What is this PR for?
Retrieve SUBMARINE_AUTH_SECRET from environment variable instead of using hard-coded value

### What type of PR is it?
Bug Fix

### What is the Jira issue?
https://issues.apache.org/jira/browse/SUBMARINE-1417
### Questions:

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -22,6 +22,7 @@ env:
   VERSION: "0.9.0-SNAPSHOT"
   BUILD_FLAG: "clean install -ntp -DskipTests -am"
   TEST_FLAG: "test -DskipRat -ntp"
+  SUBMARINE_AUTH_SECRET: "SUBMARINE_SECRET_12345678901234567890"
 jobs:
   generate-k8s-versions-array:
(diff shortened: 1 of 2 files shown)


xunliu (approved):
LGTM
@laiyousin Thank you fixed this problem, thanks.

cdmikechen:
@laiyousin @xunliu 

`submarine.auth.default.secret` is not hard-coded and can be overridden by `SUBMARINE_AUTH_DEFAULT_SECRET` in the environment variable. 
Therefore, I think this PR still needs more consideration, and there are too many problems in CI that have not been solved, and it still needs the support of other fixes to merge.

Refer to the documentation https://submarine.apache.org/docs/designDocs/wip-designs/security-implementation#authentication for more authentication details.

cdmikechen:
If there is no more to deal with this PR, I will revert this recently and fix the problem in CI first.

